PR #9, opened 2023-11-03
docs: bump the change the base url example code per recent release changes

ryannikolaidis:
It looks like we no longer support passing in a Security object and instead have the users directly assign `api_key_auth`, looks like this part of the README just needed the respective bump.


awalker4 (approved):
Thanks, good catch!

PR #13, opened 2023-11-15
Cleaning

Coniferish:
Blackens files
Fixes example code in README


awalker4 (approved):
LGTM! Per our slack convo, the linting will probably be overwritten on the next autogenerated pr, but good thing to consider as an update to the workflow.

PR #15, opened 2023-12-05
chore: create test

Coniferish:
This PR adds a test that checks if the retry logic is operating using custom retry configurations (for the sake of time when running the test). This test also requires that an `UNS_API_KEY` be set in your local environment. This can be done by exporting it in your bash session (`export UNS_API_KEY=<api_key>`).

Additionally, it adds a `.genignore` file, which is used to tell Speakeasy to _ignore_ this test file so it is not overwritten.

### DIFF
--- a/.genignore
+++ b/.genignore
@@ -2,3 +2,6 @@
 
 # ignore human-written test files
 tests/test_utils_retries.py
+
+# ignore Makefile
+Makefile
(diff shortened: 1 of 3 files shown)


Coniferish:
@awalker4 sounds good to me! I'll add the install in the make target

Coniferish:
@awalker4, added the makefile. Lmk if you think it looks okay.

awalker4 (approved):
LGTM!

PR #25, opened 2024-01-09
feat: support image block extraction

christinestraub:
This PR added logic to pass `extract_image_block_types` parameter to API correctly.

### Testing
- `unstructured-api`
```
make run-web-app
```

- `unstructured`
```
$ pip uninstall -y unstructured-client
$ git clone -b feat/support-image-block-extraction https://github.com/Unstructured-IO/unstructured-python-client.git & cd unstructured-python-client
pip install -e .
```
Run the following code
```
from unstructured.partition.api import partition_via_api

filename = "example-docs/embedded-images-tables.pdf"

elements = partition_via_api(
    filename=filename,
    api_url="http://localhost:8000/general/v0/general",
    api_key=<api-key>,
    strategy="hi_res",
    extract_image_block_types=["image", "table"],
)
print(elements)
```

### DIFF
--- a/src/unstructured_client/utils/utils.py
+++ b/src/unstructured_client/utils/utils.py
@@ -481,16 +481,20 @@ def serialize_multipart_form(media_type: str, request: dataclass) -> Tuple[str, 
                 None, marshal_json(val, field.type), "application/json"]]
             form.append(to_append)
         else:
-            field_name = field_metadata.get(
-                "field_name", field.name)
-            if isinstance(val, List):
-                for value in val:
-                    if value is None:
-                        continue
-                    form.append(
-                        [field_name + "[]", [None, _val_to_string(value)]])
+            field_name = field_metadata.get("field_name", field.name)
+            if field_name == "extract_image_block_types":
+                # Generate a string representation of the list with double quotes
+                # e.g. '["image", "table"]'
+                form.append([field_name, [None, _val_to_string(val).replace("'", '"')]])
             else:
-                form.append([field_name, [None, _val_to_string(val)]])
+                if isinstance(val, List):
+                    for value in val:
+                        if value is None:
+                            continue
+                        form.append(
+                            [field_name + "[]", [None, _val_to_string(value)]])
+                else:
+                    form.append([field_name, [None, _val_to_string(val)]])
     return media_type, None, form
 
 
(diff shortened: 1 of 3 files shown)


Coniferish:
@christinestraub, the code in this repo is ai-generated. We can't persist code changes made manually. @awalker4 is more familiar with updating this repo than I am, but I believe we have to make updates to `openapi.json` in `unstructured-api` and the ai will update the code in this repo. I see you added `extract_image_block_types` to that file last week, but I'm unsure how to trigger the updates in this repo. @awalker4 ?

christinestraub:
@Coniferish The new param  `extract_image_block_types` was added when I modified openapi.json but it did not work. 
I investigated the cause and created this PR.

src/unstructured_client/utils/utils.py
```diff
@@ -481,16 +481,20 @@ def serialize_multipart_form(media_type: str, request: dataclass) -> Tuple[str,
...
-            field_name = field_metadata.get(
-                "field_name", field.name)
-            if isinstance(val, List):
-                for value in val:
-                    if value is None:
-                        continue
-                    form.append(
-                        [field_name + "[]", [None, _val_to_string(value)]])
+            field_name = field_metadata.get("field_name", field.name)
+            if field_name == "extract_image_block_types":
```
badGarnet, line 485: based on the comment wouldn't it be more accurate to check if `val` is a list with strings? And in general don't we need to replace single quote?

src/unstructured_client/utils/utils.py
```diff
@@ -481,16 +481,20 @@ def serialize_multipart_form(media_type: str, request: dataclass) -> Tuple[str,
...
+                # e.g. '["image", "table"]'
+                form.append([field_name, [None, _val_to_string(val).replace("'", '"')]])
             else:
-                form.append([field_name, [None, _val_to_string(val)]])
+                if isinstance(val, List):
+                    for value in val:
+                        if value is None:
+                            continue
+                        form.append(
+                            [field_name + "[]", [None, _val_to_string(value)]])
```
badGarnet, line 495: so if val is any other list (given the comment above lets say a list of integers): `[1, 2]` this would add an entries:
```python
[
    ["numbers[]", [None, "1"]],
    ["numbers[]", [None, "2"]],
]
``` 
This seem like breaking pattern?

christinestraub:
As @Coniferish commented, the code in this repo is ai-generated. We can't persist code changes made manually.

Issue #30, opened 2024-01-25
Add helper logic around the server_url arg


awalker4:
We want to help users with some of the most common client issues:
* If no server_url is passed and a 401 is returned from the freemium api, they may be a SaaS user with a valid key. Suggest to the user that they may need to enter a custom url. 
* If a path is included with the custom url, strip the path rather than return a 404
* If the url uses http or has no protocol, prefix with https instead

This will be custom logic on top of the autogenerated code. Speakeasy will provide hooks to do this soon, but for now we should establish a pattern, and maintain this code whenever a pr is regenerated. I propose a separate file with some custom decorators, and a line wrapper in general.py. When prs try to strip this line, we just bring it back.

Coniferish:
TODO: create pattern for ignoring unstructured-written files
TODO: create a make command to pop open a jupyter notebook
IDEA: create make command or branch for adding the unstructured decorators after a speakeasy PR

PR #35, opened 2024-02-06
chore: suggest url if 401

Coniferish:
Final PR for closing #30 

This PR adds a decorator to the `partition` method of the `General` class to catch if a 401 error is being raised while `server_url` is `None` and add a warning message.

Testing
```
from unstructured_client.models import shared

FAKE_KEY = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"

client = UnstructuredClient(
    api_key_auth=FAKE_KEY,
)
        
filename = "_sample_docs/layout-parser-paper-fast.pdf"
        
   files=shared.Files(
        content=f.read(),
        file_name=filename,
   )

    files=files,
)

client.general.partition(req)
```

### DIFF
--- a/_test_unstructured_client/test__decorators.py
+++ b/_test_unstructured_client/test__decorators.py
@@ -1,15 +1,11 @@
-import os
 import pytest
 
 from unstructured_client import UnstructuredClient
+from unstructured_client.models import shared
+from unstructured_client.models.errors import SDKError
 
 
-def get_api_key():
-    api_key = os.getenv("UNS_API_KEY")
-    if api_key is None:
-        raise ValueError("""UNS_API_KEY environment variable not set. 
-Set it in your current shell session with `export UNS_API_KEY=<api_key>`""")
-    return api_key
+FAKE_KEY = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
 
 
 @pytest.mark.parametrize(
@@ -25,7 +21,7 @@ Set it in your current shell session with `export UNS_API_KEY=<api_key>`""")
 def test_clean_server_url_on_paid_api_url(server_url: str):
     client = UnstructuredClient(
         server_url=server_url,
-        api_key_auth=get_api_key(),
+        api_key_auth=FAKE_KEY,
     )
     assert client.general.sdk_configuration.server_url == "https://unstructured-000mock.api.unstructuredapp.io"
 
@@ -42,7 +38,7 @@ def test_clean_server_url_on_paid_api_url(server_url: str):
 def test_clean_server_url_on_localhost(server_url: str):
     client = UnstructuredClient(
         server_url=server_url,
-        api_key_auth=get_api_key(),
+        api_key_auth=FAKE_KEY,
     )
     assert client.general.sdk_configuration.server_url == "http://localhost:8000"
 
@@ -50,9 +46,10 @@ def test_clean_server_url_on_localhost(server_url: str):
 def test_clean_server_url_on_empty_string():
     client = UnstructuredClient(
         server_url="",
-        api_key_auth=get_api_key(),
+        api_key_auth=FAKE_KEY,
     )
     assert client.general.sdk_configuration.server_url == ""
+
 
 @pytest.mark.parametrize(
     ("server_url"),
@@ -63,8 +60,33 @@ def test_clean_server_url_on_empty_string():
 )
 def test_clean_server_url_with_positional_arguments(server_url: str):
     client = UnstructuredClient(
-        get_api_key(),
+        FAKE_KEY,
         "",
         server_url,
     )
     assert client.general.sdk_configuration.server_url == "https://unstructured-000mock.api.unstructuredapp.io"
+
+
+def test_suggest_defining_url_if_401():
+    with pytest.warns(UserWarning):
+
+        client = UnstructuredClient(
+            api_key_auth=FAKE_KEY,
+        )
+        
+        filename = "_sample_docs/layout-parser-paper-fast.pdf"
+        
+        with open(filename, "rb") as f:
+            files=shared.Files(
+                content=f.read(),
+                file_name=filename,
+            )
+
+        req = shared.PartitionParameters(
+            files=files,
+        )
+
+        try:
+            client.general.partition(req)
+        except SDKError as e:
+            print(e)
--- a/_test_unstructured_client/test__decorators.py
+++ b/_test_unstructured_client/test__decorators.py
@@ -51,10 +51,7 @@ def test_clean_server_url_fixes_malformed_localhost_url(server_url: str):
 
 
 def test_clean_server_url_returns_empty_string_given_empty_string():
-    client = UnstructuredClient(
-        server_url="",
-        api_key_auth=FAKE_KEY,
-    )
+    client = UnstructuredClient( server_url="", api_key_auth=FAKE_KEY)
     assert client.general.sdk_configuration.server_url == ""
 
 
--- a/_test_unstructured_client/test__decorators.py
+++ b/_test_unstructured_client/test__decorators.py
@@ -9,16 +9,18 @@ FAKE_KEY = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
 
 
 @pytest.mark.parametrize(
-    ("server_url"),
+    "server_url",
     [
-        ("https://unstructured-000mock.api.unstructuredapp.io"),  # correct url
-        ("unstructured-000mock.api.unstructuredapp.io"),
-        ("http://unstructured-000mock.api.unstructuredapp.io/general/v0/general"),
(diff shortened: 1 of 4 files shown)


_test_unstructured_client/test__decorators.py
```diff
@@ -1,70 +1,98 @@
...
-        ("https://unstructured-000mock.api.unstructuredapp.io"), # correct url
+        ("https://unstructured-000mock.api.unstructuredapp.io"),  # correct url
         ("unstructured-000mock.api.unstructuredapp.io"),
         ("http://unstructured-000mock.api.unstructuredapp.io/general/v0/general"),
         ("https://unstructured-000mock.api.unstructuredapp.io/general/v0/general"),
         ("unstructured-000mock.api.unstructuredapp.io/general/v0/general"),
-    ]
+    ],
 )
 def test_clean_server_url_on_paid_api_url(server_url: str):
```
scanny, line 21: I recommend you make test titles state the behavior being tested. In this case, something like:

```python
def test_it_replaces_a_misshapen_server_url_with_the_correct_url(...)
```
The name should be a verb-phrase, not a noun phrase.

I was able to work out roughly what this was doing by reading the code, but that's unnecessary work/time and is error prone. It's good to be able to read just the test names and get a good idea of behaviors the unit-under-test has. That can also help with identifying missing tests. A unit should do this and do that. It does not be this and be that.

A good title also helps a reader (importantly including yourself) to reason about whether all the necessary cases are being tested.

I know you didn't name this test in this PR but if you touch it you own it and you should feel free to remedy any shortcomings left by the last developer.

_test_unstructured_client/test__decorators.py
```diff
@@ -1,70 +1,98 @@
...
+    )
+    assert (
+        client.general.sdk_configuration.server_url
+        == "https://unstructured-000mock.api.unstructuredapp.io"
     )
-    assert client.general.sdk_configuration.server_url == "https://unstructured-000mock.api.unstructuredapp.io"
 
 
 @pytest.mark.parametrize(
     ("server_url"),
```
scanny, line 33: This should either be `str` (single param) or `tuple[str]` (for multiple params). So either remove the parens or add a trailing comma.

_test_unstructured_client/test__decorators.py
```diff
@@ -1,70 +1,98 @@
...
     [
-        ("http://localhost:8000"), # correct url
+        ("http://localhost:8000"),  # correct url
```
scanny, line 35: On these kinds of comments, I'd be inclined to make this a "header" comment rather than a trailing one, and add a heading comment saying what the others are in contrast. Something like:
```python
# -- well-formed url --
"http://localhost:8000",
# -- url commonly ill-formed by users --
"localhost:8000",
...
```

Also note the parens are not necessary here since there is only the single parameter so those should come out. I'm surprised Black didn't do that automatically, maybe it's not running here.

_test_unstructured_client/test__decorators.py
```diff
@@ -1,70 +1,98 @@
...
         server_url,
     )
-    assert client.general.sdk_configuration.server_url == "https://unstructured-000mock.api.unstructuredapp.io"
+    assert (
+        client.general.sdk_configuration.server_url
+        == "https://unstructured-000mock.api.unstructuredapp.io"
+    )
+
+
+def test_suggest_defining_url_if_401():
```
scanny, line 76: This name is better. I find it helpful to add the `it_...` prefix to get the descriptive ball rolling. I would mention how it suggests because that's not typically a REST API can do (as opposed to a GUI). Maybe `test_it_issues_a_warning_to_suggest_defining_a_URL_on_a_401()`.

_test_unstructured_client/test__decorators.py
```diff
@@ -1,70 +1,98 @@
...
+            )
+
+        req = shared.PartitionParameters(
+            files=files,
+        )
+
+        try:
+            client.general.partition(req)
+        except SDKError as e:
+            print(e)
```
scanny, line 98: What is the part that's supposed to issue a warning?

Tighten your `with pytest.warns()` context to the line that is supposed to do that. Otherwise the reader has no indication of what's triggering the warning.

Also, assert the warning message is what you expect. This is an example of doing that with `pytest.raises()`, I expect the same is available for `pytest.warns()`: https://github.com/Unstructured-IO/unstructured/blob/main/test_unstructured/partition/docx/test_docx.py#L282

This even more vividly describes the behavior to the reader.

scanny, line 98: Why is there a `print()` statement here?

src/unstructured_client/utils/_decorators.py
```diff
@@ -39,8 +45,29 @@ def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> None:
...
+                args = args[:SERVER_URL_ARG_IDX] + (urlunparse(cleaned_url),) + args[SERVER_URL_ARG_IDX + 1 :]  # type: ignore
+
         return func(*args, **kwargs)
 
     return wrapper
+
+
+def suggest_defining_url_if_401(
+    func: Callable[_P, operations.PartitionResponse]
+) -> Callable[_P, operations.PartitionResponse]:
```
scanny, line 57: Every module, class, method, and function should have a docstring. Use it to articulate for yourself what the contract of that method or function is or what the role of a module or class is. All docstrings should be PEP 257 compliant, like:
```python
def func(...):
    """Concise single-line statement of contract.
{blank line}
    More detail when necessary to do things like describe when a function with an
    Optional[..] return value returns None instead of that value, etc.
    """
```

This particular decorator has plenty worth saying about it, like that it's used to patch generated code, needs to be applied by hand every time, what exactly if fixes about the generated code, etc. Document your work for others who follow.

scanny (changes requested):
A few remarks. Code looks fine to me.

_test_unstructured_client/test__decorators.py
Coniferish: I grabbed the bases of this from the README example, which had the print statement. I'll remove it

_test_unstructured_client/test__decorators.py
```diff
@@ -0,0 +1,113 @@
...
+        "http://unstructured-000mock.api.unstructuredapp.io/general/v0/general",
+        "https://unstructured-000mock.api.unstructuredapp.io/general/v0/general",
+        "unstructured-000mock.api.unstructuredapp.io/general/v0/general",
+    ],
+)
+def test_clean_server_url_fixes_malformed_paid_api_url(server_url: str):
+    client = UnstructuredClient(
+        server_url=server_url,
+        api_key_auth=FAKE_KEY,
+    )
```
scanny, line 27: These will all fit on one line, won't they? Vertical space is something to preserve on principle. Leaving off the trailing comma on the last argument will keep black from automatically formatting it this way. Each of these tests uses up more space than necessary.

src/unstructured_client/general.py
```diff
@@ -12,7 +14,7 @@ def __init__(self, sdk_config: SDKConfiguration) -> None:
         self.sdk_configuration = sdk_config
         
     
-    
+    @suggest_defining_url_if_401 # human code
```
scanny, line 17: This isn't being linted, I can tell because there's only one space in front of the `#` for the comment. Isn't there a lint step in the CI process?

src/unstructured_client/utils/_decorators.py
```diff
@@ -1,15 +1,28 @@
...
 _P = ParamSpec("_P")
 
 
 def clean_server_url(func: Callable[_P, None]) -> Callable[_P, None]:
+    """A decorator for fixing common types of malformed 'server_url' arguments.
+
+    This decorator addresses the common problem of users omitting or using the wrong url scheme and/or
+    adding the '/general/v0/general' path to the 'server_url'.
+    The decorator should be manually applied to the __init__ method of UnstructuredClient after merging
+    a PR from Speakeasy.
```
scanny, line 24: Make this one paragraph or two, but put a blank line between paragraphs. I think one paragraph is plenty here.

Look for a word-wrap setting/keystroke in your editor to speed up reformatting between the margins.

scanny (approved):
LGTM other than a couple formatting issues I marked. Approving in advance.

_test_unstructured_client/test__decorators.py
Coniferish: That was made from using black, but I can revert it

src/unstructured_client/general.py
Coniferish: Not on this repo. I am manually blackening the files I missed this one I guess.

Coniferish: Actually, the reason I didn't lint this one is because it's one of Speakeasy's and any linting on this file would be overwritten the next time Speakeasy made a PR.

PR #39, opened 2024-02-09
chore: log retry

### LINKED ISSUE #31: Add visibility to the client retries
When the client hits a retryable error, we should log the status code, error message, and time until next try. Similar to #30 , we're wrapping the generated code for now. Need to decide how we want to hook into `retry_with_backoff` to do this. Speakeasy will address this soon, so this is just to tie us over.

Coniferish:
Closes #31 
Adds function to log retries. A function is used instead of a decorator because it is not possible to use a decorator to access the `while` loop in `retry_with_backoff` where retries are made.

Also renames `_decorators.py` to `_human_utils.py`

In `_human_utils.py`, `from unstructured_client.models import errors` is moved to prevent a circular import error between the `models` and `utils` modules.

### Testing
- from your unstructured venv, `pip uninstall unstructured-client`
- navigate to unstructured-python-client, check out to this branch, and `pip install -e .`
- `pip install requests_mock`
- open a python shell and run the following:
```
from unstructured.partition.api import partition_via_api
import requests_mock

with requests_mock.Mocker() as mock:
    mock.post("https://api.unstructured.io/general/v0/general", status_code=500)
    elements = partition_via_api(filename="example-docs/fake-email.msg")
```

### DIFF
--- a/_test_unstructured_client/test_utils_retries.py
+++ b/_test_unstructured_client/test_utils_retries.py
@@ -1,5 +1,6 @@
 import pytest
 import logging
+import re
 
 import requests_mock
 
@@ -27,10 +28,7 @@ def test_retry_with_backoff_does_retry(caplog):
         session = UnstructuredClient(api_key_auth=FAKE_KEY)
 
         with open(filename, "rb") as f:
-            files=shared.Files(
-                content=f.read(),
-                file_name=filename,
-            )
+            files=shared.Files(content=f.read(), file_name=filename)
 
         req = shared.PartitionParameters(files=files)
 
@@ -59,14 +57,10 @@ def test_backoff_strategy_logs_retries(caplog):
         session = UnstructuredClient(api_key_auth=FAKE_KEY)
 
         with open(filename, "rb") as f:
-            files=shared.Files(
-                content=f.read(),
-                file_name=filename,
-            )
+            files=shared.Files(content=f.read(), file_name=filename)
 
         req = shared.PartitionParameters(files=files)
         with pytest.raises(Exception):
             session.general.partition(req, retries=retries)    
-
-    
-    assert "seconds before retry" in caplog.text
+    pattern = re.compile(f"{re.escape('Retry attempt #1. Sleeping')}.*{'seconds before retry'}")
+    assert bool(pattern.search(caplog.text))
--- a/_test_unstructured_client/test_utils_retries.py
+++ b/_test_unstructured_client/test_utils_retries.py
@@ -1,5 +1,5 @@
-import os
 import pytest
+import logging
 
 import requests_mock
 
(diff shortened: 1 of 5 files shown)


_test_unstructured_client/test_utils_retries.py
```diff
@@ -45,3 +41,32 @@ def test_backoff_strategy():
...
+    with requests_mock.Mocker() as mock:
+        # mock a 500 status code for POST requests to the api 
+        mock.post("https://api.unstructured.io/general/v0/general", status_code=500)
+        session = UnstructuredClient(api_key_auth=FAKE_KEY)
+
+        with open(filename, "rb") as f:
+            files=shared.Files(
+                content=f.read(),
+                file_name=filename,
+            )
```
scanny, line 65: This will all fit on one line, won't it?

_test_unstructured_client/test_utils_retries.py
```diff
@@ -45,3 +41,32 @@ def test_backoff_strategy():
...
+
+        req = shared.PartitionParameters(files=files)
+        with pytest.raises(Exception):
+            session.general.partition(req, retries=retries)    
+
+    
+    assert "seconds before retry" in caplog.text
```
scanny, line 72: This assert looks a little skinny. Can you include more of the exact text that will appear? As a reader, I'd like to have a sense of the user experience when this happens, like a line appears in the log that says "Retrying request ... " or whatever they're going to actually see. Best case would be showing the entire and exact log contents. Does it retry more than once for example? You can make the retry interval very short so it doesn't produce an unnecessarily long test duration.

scanny (approved):
A couple remarks, otherwise LGTM.

_test_unstructured_client/test_utils_retries.py
Coniferish: I can make it longer. The log just isn't the same every time because the interval between retries is variable (we randomize it to an extent). 
Yah, it does retry more than once. This test uses a customized backoff strategy so that it retries a few times, but not for longer than 30 sec or something.

scanny: One thing you might consider is filtering the log if it contains other bits and bobs, like:
```python
lines_I_care_about = [line for line in caplog.text.splitlines() if "identifying text" in line]
```
Then you could assert on the count which assures the number of retries you expect, etc.

Coniferish: There's another test that tests retries occurs, so I was thinking of just keeping this one to just checking that logging is happening.

src/unstructured_client/utils/_human_utils.py
```diff
@@ -88,3 +91,13 @@ def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> operations.PartitionResponse
...
     return wrapper
+
+
+def log_retries(retry_count: int, sleep: float):
+    """Function for logging retries to give users visibility into requests."""
+    logging.basicConfig(level=logging.INFO, format='%(levelname)s: %(message)s', stream=sys.stdout)
+    logger = logging.getLogger('unstructured-client')
+    logger.setLevel(logging.INFO)
+    logger.info(
+        "Retry attempt #%s. Sleeping %s seconds before retry.", retry_count, round(sleep, 1)
```
awalker4, line 102: Any chance we can access the response here? Would be great to also dump the status code and any error message that came back.

hubert-rutkowski85:
Hey @Coniferish @awalker4 I propose to squash branch before merging to main, there should be checkbox to do that automatically.

image

I see other repos follow this, which gives better overview on main of what happened recently, without going into individual commit details:

image

In both cases we see 9 days worth of history :)

Coniferish:
Thanks @hubert-rutkowski85. I was just used to those setting existing in the other repos. They aren't set up on this one because it is ai generated. I'll look into setting that up for us, though.

PR #44, opened 2024-02-17
Revert "Move the openapi source of truth"

awalker4:
This reverts commit ba95fb0f0adfbc374d320b31e18a672190b90a18. See https://github.com/Unstructured-IO/unstructured-api/pull/369

### DIFF
--- a/.github/workflows/speakeasy_sdk_generation.yml
+++ b/.github/workflows/speakeasy_sdk_generation.yml
@@ -24,7 +24,7 @@ jobs:
       create_release: true
       openapi_doc_auth_header: x-api-key
       openapi_docs: |
-        - https://api.unstructured.io/general/openapi.json
+        - https://raw.githubusercontent.com/Unstructured-IO/unstructured-api/main/openapi.json
       publish_python: true
       speakeasy_version: latest
     secrets:


.github/workflows/speakeasy_sdk_generation.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       create_release: true
       openapi_doc_auth_header: x-api-key
       openapi_docs: |
-        - https://api.unstructured.io/general/openapi.json
+        - https://raw.githubusercontent.com/Unstructured-IO/unstructured-api/main/openapi.json
```
hubert-rutkowski85, line 27: We don't want to point at the file in repo?  https://github.com/marketplace/actions/speakeasy-client-sdk-generation-action#openapi_docs

```
      openapi_docs: |
        - ../../openapi.json
```
Spares us having to manually update it every time something changes (and risking forgetting).

awalker4: Hopefully should work now that https://github.com/Unstructured-IO/unstructured-api/pull/369 is merged. My thinking is that the source of truth for params should live with the server code, and we'll update whenever we add a new param in unstructured-api. (It will allow both clients to use the same source.) Although, I didn't realize we could supply multiple files - I wonder if we want to keep client specific params in a local `openapi.json` and let them merge, rather than having to generate a special overlay file.

Coniferish:
Tagging this related PR: https://github.com/Unstructured-IO/unstructured-js-client/pull/28
I merged before seeing the conversation here. Depending on the outcome, the same changes should be applied to the js client repo

.github/workflows/speakeasy_sdk_generation.yml
hubert-rutkowski85: Yes, url is 200 now :) I also think that unstructured-api should contain source of truth for clients, but we can't use it now directly as there are too big changes in client from it. So we will have to update the openapi.json in this repo whenever backend changes for some time. My point was that we could use the file from repo because I thought the link points to manually uploaded file. I mistaken it for github gists, but it actually points to the file in repo. So it's perfect :) Approving.

Re overlays: because there are big differences between server and client openapis, we probably can't do direct comparison like I planned here: https://github.com/Unstructured-IO/unstructured-python-client/pull/41/commits/579f2ebb3df54724408db7ba6fa7f07387aa3ac9 Then overlays don't help but add complexity. In such case for simplicity I think to just have a single openapi.json with hand edited changes, plus a well documented .txt doc with list of those changes.

Re js client: didn't look inside yet. Will look after finishing pdf splitting.

PR #47, opened 2024-02-23
Jj/pr configs

### LINKED ISSUE #43: chore: add PR conventions
Unlike the other unstructured repos, this one does not have CI/CD or other standardizing scripts set up. 

- Add template for PRs and opening Issues
- Add workflows for running tests before merging PRs

Coniferish:
Closes #43

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,48 @@
+name: CI
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  merge_group:
+    branches: [ main ]
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  setup:
+    strategy:
+      matrix:
+        python-version: ["3.9","3.10","3.11"]
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: ./.github/actions/base-cache
+      with:
+        python-version: ${{ matrix.python-version }}
+        check-only: 'true'
+
+  test_unit:
+    strategy:
+      matrix:
+        python-version: ["3.9","3.10","3.11"]
+    runs-on: ubuntu-latest
+    needs: [setup]
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Setup virtual environment
+      uses: ./.github/actions/base-cache
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Test
+      run: |
+        source .venv/bin/activate
+        sudo apt-get update
+        make test CI=true UNSTRUCTURED_INCLUDE_DEBUG_METADATA=true
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,27 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: bug/<short-name>
+labels: bug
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: make install-test
+    - name: Run unit tests
+      run: |
+        pip install -e .
```
rbiseck3, line 31: Does this need to be installed in editable mode? 
```suggestion
        pip install  .
```

Coniferish: No. That was just my default for installing.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,32 @@
...
+    branches: [ main ]
+  merge_group:
+    branches: [ main ]
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  test_unit:
```
hubert-rutkowski85, line 16: It's probably worth adding a separate job for linting.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: make install-test
+    - name: Run unit tests
+      run: |
+        pip install  .
+        make test
```
hubert-rutkowski85, line 32: When https://github.com/Unstructured-IO/unstructured-python-client/pull/41 is merged, the `make test` will run also integration tests, which means docker. It would be good if that worked here just like that, but if not, then running `make test-unit` will be simple solution.

Coniferish: I can make that change once it's merged, but #41, which adds `make test-unit`, also hasn't been merged so it doesn't make sense to update this yet.

hubert-rutkowski85: That's right. I thought it will be merged faster, but this one's ready, so accepting :)

.github/workflows/ci.yaml
```diff
@@ -30,3 +30,12 @@ jobs:
...
         make test
+
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install dependencies
+      run: pip install .[dev]
+    - name: Lint
+      run: pylint --rcfile=pylintrc src
```
hubert-rutkowski85, line 41: Hmm, why not `make lint` ?

.github/workflows/ci.yaml
```diff
@@ -30,3 +30,12 @@ jobs:
       run: |
         pip install  .
         make test
+
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install dependencies
+      run: pip install .[dev]
```
hubert-rutkowski85, line 39: I see pylint mentioned in `make install-dev` , and in `extras_require` in setup.py . It looks like two alternative methods to install a single library. We should probably standardize on one way. I'm not sure yet on which, so also not requesting any changes here. Just pointing it out loud, so everyone knows that could be improved in future.

hubert-rutkowski85 (approved):

.github/workflows/ci.yaml
Coniferish: I'll try that again, but I think it was raising an error because it wasn't being called in the project directory

Coniferish: Looks like I misread the error before and it works fine :)

PR #64, opened 2024-03-27
Move decorator logic to hooks

### LINKED ISSUE #60: chore/custom client logic is moved to hooks
**Definition of done**
* The custom logic for pdf splitting, retry logging, and server_url fixing are moved to the appropriate lifecycle hook
* All of this logic is tested locally and still works
* Client is regenerated without needing to restore any human written code

Coniferish:
UPDATE: This is a work in progress and I will be migrating the additional `handle_split_pdf_page` decorator in this hook.

Closes #60
-  The custom logic for pdf splitting, retry logging, and server_url fixing are moved to the appropriate lifecycle hook
-  All of this logic is tested locally and still works (can be verified with `make test`)
-  Client is regenerated without needing to restore any human written code (with the exception of the `handle_split_pdf_page` decorator).
   - You can checkout to this branch and run `make client-generate` and no changes should be staged
   
It's recommended to review commit by commit as the first commit was generated by speakeasy


src/unstructured_client/_hooks/registration.py
```diff
@@ -1,13 +1,98 @@
...
 
 
 # This file is only ever generated once on the first generation and then is free to be modified.
-# Any hooks you wish to add should be registered in the init_hooks function. Feel free to define them
-# in this file or in separate files in the hooks folder.
+# Any hooks you wish to add should be registered in the init_hooks function. Feel free to define 
+# them in this file or in separate files in the hooks folder.
+
+
+class CleanServerUrlSDKInitHook(SDKInitHook):
```
scanny, line 17: Every new class (and method and module) should have a docstring that orients the new reader to what the role of the class or module is and how it fits into the bigger picture. I'd say this module and these classes are no exception to that.

scanny (approved):
Yep, LGTM. One request for documentation.

I'm not super clear how all this works so I assume you know what you're doing, but overall it seems like we're replacing the "patch-it-in-every-time" work we used to have to do with one-time hooks which survive regeneration. So that's a great step forward :)

Approving in advance.

PR #69, opened 2024-04-22
Move Split PDF page functionality to hooks

mpolomdeepsense:
<b> ONLY `_hooks` directory and `gen.yaml` file are relevant for review. Rest of the changes were generated by speakeasy </b>


mpolomdeepsense:
@hubert-rutkowski85 @awalker4 

PR is still missing some unit tests, I'm working on it right now, but the main functionality is finished and can be reviewed.

src/unstructured_client/_hooks/custom/clean_server_url_hook.py
```diff
@@ -0,0 +1,34 @@
+from typing import Tuple
+from urllib.parse import ParseResult, urlparse, urlunparse
+
+import requests
+
+from ..types import SDKInitHook
```
hubert-rutkowski85, line 6: It's better to avoid using relative imports..

src/unstructured_client/_hooks/custom/log_retries_after_error_hook.py
```diff
@@ -0,0 +1,42 @@
+import logging
+import sys
+from typing import Optional, Tuple, Union
+
+import requests
+
+from ..types import AfterErrorContext, AfterErrorHook, SDKInitHook
```
hubert-rutkowski85, line 7: Ditto

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -0,0 +1,505 @@
...
+                    executor.submit(call_api_partial, (page_content, page_number))
+                )
+                if page_number == all_pages_number - 1:
+                    break
+
+        last_page_request = self._create_request(
+            request, form_data, next(pages)[:2], filename
+        )
+        last_page_prepared_request = self.client.prepare_request(last_page_request)
+        return last_page_prepared_request
```
hubert-rutkowski85, line 149: I assume the division of requests to have the last page sent separately, is needed to make it work with the hooks way of work, because it needs to return some request. It would be good to point it out explicitely for reader in comment.

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -0,0 +1,505 @@
...
+            or not file.filename.endswith(".pdf")
+        ):
+            logger.warning("Given file is not a PDF. Continuing without splitting.")
+            return request
+
+        if self.client is None:
+            logger.warning("HTTP client not accessible! Continuing without splitting.")
+            return request
+
+        filename = file.filename.replace(".pdf", "")
```
hubert-rutkowski85, line 127: It would be good to add comment why you're removing .pdf extension.

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -0,0 +1,505 @@
...
+        with ThreadPoolExecutor(max_workers=call_threads) as executor:
+            for page_content, page_number, all_pages_number in pages:
+                self.partition_requests[operation_id].append(
+                    executor.submit(call_api_partial, (page_content, page_number))
+                )
+                if page_number == all_pages_number - 1:
+                    break
+
+        last_page_request = self._create_request(
+            request, form_data, next(pages)[:2], filename
```
hubert-rutkowski85, line 146: The `next(page)[:2]` looks a bit cryptic on first glance, maybe there's a nicer way to extract page_content and page_number, eg
```
last_page = next(pages)
last_page_content, last_page_number = last_page[:2]
``

File: src/unstructured_client/_hooks/custom/split_pdf_hook.py [hunk 6]
```diff
@@ -0,0 +1,505 @@
...
+            part_params = self._decode_content_disposition(content_disposition)
+            name = part_params.get("name")
+
+            if name is None:
+                continue
+
+            if name == PARTITION_FORM_FILES_KEY:
+                filename = part_params.get("filename") or SUBSTITUTE_FILENAME
+                form_data[PARTITION_FORM_FILES_KEY] = File(
+                    filename.split("/")[-1], part.content
```
hubert-rutkowski85, line 280: Does it assume that a filename will have at least one `/` inside? But user can pass a file with `name.pdf`

src/unstructured_client/_hooks/custom/split_pdf_hook.py
hubert-rutkowski85: Also, the last page is not put into `self.partition_requests` - I assume it's ok? The _await_elements doesn't use it? Couldn't see in code where it is picked up.

mpolomdeepsense: It's passed to `_await_elements` as a response argument in `after_success` or `after_error`:

at the bottom of the `_await_elements` function 
```python
if response.status_code == 200:
    elements.append(response.json())
```

src/unstructured_client/_hooks/custom/log_retries_after_error_hook.py
mpolomdeepsense: Fixed

src/unstructured_client/_hooks/custom/clean_server_url_hook.py
mpolomdeepsense: Yep, forgot to change that. Fixed.

src/unstructured_client/_hooks/custom/split_pdf_hook.py
mpolomdeepsense: Done

src/unstructured_client/_hooks/custom/split_pdf_hook.py
mpolomdeepsense: Added comment here and to the `after_success` and `after_error` methods

src/unstructured_client/_hooks/custom/split_pdf_hook.py
mpolomdeepsense: Changed as suggested

src/unstructured_client/_hooks/custom/split_pdf_hook.py
mpolomdeepsense: Actually changed it so it doesn't add the page number to the file name. This way the matadata is more consistent between split and non split pdf calls.

src/unstructured_client/_hooks/custom/split_pdf_hook.py
hubert-rutkowski85: Ah, unfortunately it's clever: https://www.joshwcomeau.com/career/clever-code-considered-harmful/#the-intern-test-1

_test_unstructured_client/test__decorators.py
```diff
@@ -81,6 +80,13 @@ def test_integration_split_pdf_has_same_output_as_non_split(
...
     # Difference in the parent_id is expected, because parent_ids are assigned when element crosses page boundary
-    diff = DeepDiff(t1=resp_split.elements, t2=resp_single.elements,
-                    exclude_regex_paths=r"root\[\d+\]\['metadata'\]\['parent_id'\]")
+    diff = DeepDiff(
+        t1=resp_split.elements,
+        t2=resp_single.elements,
+        exclude_regex_paths=[
+            r"root\[\d+\]\['metadata'\]\['parent_id'\]",
+            # TODO: (Marek Połom) - Remove page number pattern after page numbering parameter is added
+            r"root\[\d+\]\['metadata'\]\['page_number'\]",
```
hubert-rutkowski85, line 89: Do you know when will the page numbering parameter be added?

_test_unstructured_client/test_split_pdf_hook.py
```diff
@@ -0,0 +1,268 @@
...
+class TestSplitPdfHook(TestCase):
+
+    def test_unit_sdk_init(self):
+        hook = SplitPdfHook()
+        test_url = "http://localhost:5000"
+        test_client = requests.Session()
+
+        hook.sdk_init(test_url, test_client)
+
+        self.assertEqual(hook.client, test_client)
```
hubert-rutkowski85, line 25: Very nice class and all the tests! Just one question, does it require the API to be really running on port 5000 (then it's more like integration than unit test)?

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -0,0 +1,529 @@
...
+        with ThreadPoolExecutor(max_workers=call_threads) as executor:
+            for page_content, page_number, all_pages_number in pages:
+                self.partition_requests[operation_id].append(
+                    executor.submit(call_api_partial, (page_content, page_number))
+                )
+                # Check if the next page will be the last one
+                if page_number == all_pages_number - 1:
+                    break
+
+        # Before request hook needs to return a request so we skip sending the last page in parallel
```
hubert-rutkowski85, line 129: Didn't you want to write here "Because request hook" instead?

_test_unstructured_client/test_split_pdf_hook.py
mpolomdeepsense: Nope, it doesn't. You suggest to change it to some non url like string?

src/unstructured_client/_hooks/custom/split_pdf_hook.py
mpolomdeepsense: Good catch. It's the third option actually: `before_request`.

_test_unstructured_client/test__decorators.py
hubert-rutkowski85: Ok - then it would be good idea to add here CORE number or link, so it's easy to find what the comment talks about

_test_unstructured_client/test_split_pdf_hook.py
mpolomdeepsense: Added

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -0,0 +1,530 @@
...
+
+        Args:
+            file (File): The file to be checked.
+
+        Returns:
+            bool: True if the file is a PDF, False otherwise.
+        """
+        if not file.file_name.endswith(".pdf"):
+            logger.warning("Given file is not a PDF. Continuing without splitting.")
+            return False
```
hubert-rutkowski85, line 224: I think there is one more case, when file_name is empty - user just inserts the file from memory, then there is no extension to check. The discussion about how to handle pdf recognition is here https://github.com/Unstructured-IO/unstructured-python-client/pull/41#pullrequestreview-1891712616

hubert-rutkowski85: So your code will fail for memory-inserted files. If there is no filename, then reading the pdf as below has to be done.

src/unstructured_client/_hooks/custom/split_pdf_hook.py
mpolomdeepsense: Just checked and there is actually one case when it could possibly be an empty list: `"".split()`.

Added a condition that will handle this edge case.

src/unstructured_client/_hooks/custom/split_pdf_hook.py
mpolomdeepsense: Method `_parse_form_data` will use `SUBSTITUTE_FILENAME` if there is no filename or when it's an empty string. So this won't happen.

Input of this method is taken from what `_parse_form_data` returns.

hubert-rutkowski85: Hmm so setting a susbtitute filename like "foobar.pdf" is akin to cheating, as it gives it .pdf extension and we really don't know at this point what is the file type. That could cause some other code in future to work bad, as it expects to have empty filename (and perhaps different content type). So if you can refactor it now so that it works without overriding filename, that would be better. But if it's hard/timeconsuming and you want to close this PR soon, then we can improve it in future.

hubert-rutkowski85 (approved):
LGTM

src/unstructured_client/_hooks/custom/split_pdf_hook.py
mpolomdeepsense: Changed

mpolomdeepsense:
@awalker4

FYI Everything with this PR is ready from my side. If there are no comments from your side then I think it can be safely merged.

awalker4 (approved):
Sorry for the delay!

PR #77, opened 2024-05-06
update url in make command

Coniferish:
Update the url used in the `client-generate` make command


awalker4 (approved):

awalker4:
Tests were already broken - merging this while we fix those.

PR #86, opened 2024-05-14
Parametrize page splitting logic with num of threads, min and max pages per thread

micmarty-deepsense:


src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -242,9 +250,18 @@ def _is_pdf(self, file: shared.Files) -> bool:
 
         return True
 
+    def _get_optimal_split_size(self, num_pages: int, num_threads: int) -> int:
```
micmarty-deepsense, line 253: here's the relevant part

test_unstructured_client/integration/test_decorators.py
```diff
@@ -12,20 +12,22 @@
...
-    "filename, expected_ok",
+    ("filename", "expected_ok", "strategy"),
     [
-        ("_sample_docs/list-item-example-1.pdf", True),  # 1 page
-        ("_sample_docs/layout-parser-paper-fast.pdf", True),  # 2 pages
-        ("_sample_docs/layout-parser-paper.pdf", True),  # 16 pages
-        ("_sample_docs/fake.doc", True),
+        ("_sample_docs/list-item-example-1.pdf", True, "fast"),  # 1 page
+        ("_sample_docs/layout-parser-paper-fast.pdf", True, "fast"),  # 2 pages
+        # NOTE(mike): using "fast" strategy fails on this file for unknown reasons
```
micmarty-deepsense, line 19: see this post on Slack: https://unstructuredai.slack.com/archives/C04TKD85KC6/p1715777702633619

Changing "fast" to "hi_res" fixed the issue **(which deserves a separate PR by the way❗)**

test_unstructured_client/unit/test_split_pdf_hook.py
```diff
@@ -342,3 +352,63 @@ def test_unit_get_starting_page_number_missing_key(self):
...
+        description = "Large thread count for a small PDF"
+        num_pages = 3
+        num_threads = 10
+        expected_split_size = 2
+        split_size = SplitPdfHook()._get_optimal_split_size(num_pages, num_threads)
+        self.assertEqual(
+            split_size,
+            expected_split_size,
+            f"{description} => Expected: {expected_split_size}, Got: {split_size}",
+        )
```
micmarty-deepsense, line 386: all of this could be more compact using pytest.parametrize, but on the other hand, this class seems to be a good place for testing such things

overlay_client.yaml
```diff
@@ -5,4 +5,19 @@ info:
...
+        }
+  - target: $["components"]["schemas"]["partition_parameters"]["properties"]
+    update:
+      "split_pdf_threads":
+        {
+          "title": "Split Pdf Threads",
+          "description": "Number of threads used when sending requests of splitted PDF. Ignored on backend.",
+          "type": "integer",
+          "default": 5,
+        }
```
badGarnet, line 23: could you explain what does "ignored on backend" means? Does it mean splitting is just on client side and this is not passed to the server?

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -5,26 +5,26 @@
 import io
 import json
 import logging
-import os
-from concurrent.futures import ThreadPoolExecutor, Future
-from typing import Optional, Tuple, Union, Generator
+import math
+from concurrent.futures import Future, ThreadPoolExecutor
```
badGarnet, line 9: this could just be me overthinking:
- threadpool can trip on GIL and lead to unintended thread lock
- process pool provides process isolation
- we don't actually share data between threads so there is no advantage to have them under one process
- so a process pool would be safer option here

.genignore
```diff
@@ -4,7 +4,7 @@
 src/unstructured_client/_unstructured
 _jupyter
 _sample_docs
-_test_unstructured_client
```
Coniferish, line 7: What was the reason for the file name change? I'm open to changing it, but I found the convention of using a leading underscore helpful for organizing the human-written code that we maintain.

src/unstructured_client/_hooks/custom/split_pdf_hook.py
micmarty-deepsense: yeah, I'll change it to `ProcessPoolExecutor`

tests/helpers.py
```diff
@@ -0,0 +1,61 @@
...
+
+    query = parts[1]
+    params = query.split("&")
+
+    params.sort(key=lambda x: x.split('=')[0])
+
+    return parts[0] + "?" + "&".join(params)
+
+
+def sort_serialized_maps(inp: any, regex: str, delim: str):
```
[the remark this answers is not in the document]
micmarty-deepsense: At the top of this file there's the following comment:
> """Code generated by Speakeasy (https://speakeasyapi.dev). DO NOT EDIT."""

README.md
```diff
@@ -113,8 +113,8 @@ as single result. This will work only for PDF files, so don't set it for other f
 Warning: this feature causes the `parent_id` metadata generation in elements to be disabled, as that
 requires having context of multiple pages.
 
-The amount of threads that will be used for sending individual pdf pages, is controlled by
-`UNSTRUCTURED_CLIENT_SPLIT_CALL_THREADS` env var. By default it equals to 5. 
+The amount of workers that will be used for sending individual pdf pages, is controlled by
+`split_pdf_concurrency_level`. By default it equals to 5.
```
mpolomdeepsense, line 117: I also added info about possible split sizes to js-client, might be useful to add this here too.

> Size of each batch is determined internally and it can vary between 2 and 20 pages per split.

README.md
```diff
@@ -113,8 +113,8 @@ as single result. This will work only for PDF files, so don't set it for other f
 Warning: this feature causes the `parent_id` metadata generation in elements to be disabled, as that
```
mpolomdeepsense, line 113: I would also add information that split is done by default and can be disabled by setting `split_pdf_page` to `false`.

.github/workflows/ci.yaml
```diff
@@ -53,10 +53,10 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: make install-test
-    - name: Run unit tests
+    - name: Run all tests
       run: |
         pip install  .
-        pytest _test_unstructured_client/test_integration*.py
+        make test
```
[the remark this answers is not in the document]
micmarty-deepsense: yes, let's address this in a separate PR

README.md
micmarty-deepsense: done

README.md
micmarty-deepsense: done

overlay_client.yaml
micmarty-deepsense: I have updated the descriptions for both parameters ;)

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -581,3 +557,49 @@ def _get_starting_page_number(self, form_data: FormData) -> int:
             starting_page_number = DEFAULT_STARTING_PAGE_NUMBER
 
         return starting_page_number
+
+    def _get_split_pdf_call_threads(self, form_data: FormData) -> int:
```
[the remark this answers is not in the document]
micmarty-deepsense: note, that now the name is `concurrency_level`, not `num_threads` since we're using `ProcesPoolExecutor`, not `ThreadPoolExecutor`

test_unstructured_client/integration/test_decorators.py
micmarty-deepsense: done, removed

test_unstructured_client/unit/test_split_pdf_hook.py
micmarty-deepsense: done ✅

README.md
```diff
@@ -106,16 +106,16 @@ See the [general partition](/docs/models/shared/partitionparameters.md) page for
 
 #### Splitting PDF by pages
 
-In order to speed up processing of long PDF files, set `split_pdf_page=True`. It will cause the PDF
-to be split page-by-page at client side, before sending to API, and combining individual responses
+In order to speed up processing of long PDF files, `split_pdf_page=True` by default. It will cause the PDF
+to be split at client side, before sending to API, and combining individual responses
 as single result. This will work only for PDF files, so don't set it for other filetypes.
```
hubert-rutkowski85, line 111: Hmm the first sentence is a bit ungrammatical. Also, the last warning "don't set it for other filetypes" is a bit scary, because if by default we set it to true, then we shouldn't demand that user disables this flag. Maybe change of tone to "this will affect only PDF files, no need to disable it for other filetypes."

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -33,8 +33,13 @@
 PARTITION_FORM_FILES_KEY = "files"
 PARTITION_FORM_SPLIT_PDF_PAGE_KEY = "split_pdf_page"
 PARTITION_FORM_STARTING_PAGE_NUMBER_KEY = "starting_page_number"
+PARTITION_FORM_CONCURRENCY_LEVEL_KEY = "split_pdf_concurrency_level"
 
 DEFAULT_STARTING_PAGE_NUMBER = 1
+DEFAULT_CONCURRENCY_LEVEL = 5
```
Coniferish, line 39: IMO, I'd like to stick with just "threads" or "concurrency_level". This part is an example of it being slightly confusing to use both terms. 
I personally think "threads" is better/more simple. I also think `split_pdf_concurrency_level` could just be called `threads` or `call_threads`. 
One thing I'm not clear on, though, is if this argument _only_ applies to pdf documents or if it'd also apply to images or other docs.

micmarty-deepsense: Yeah, I'd also like to choose one name and stick with it. To me "concurrency level" should be clear enough for users. Using "thread" would be a lie since splits are handled by separate processes.

micmarty-deepsense: @badGarnet @awalker4 @hubert-rutkowski85 @mpolomdeepsense  what's your opinion? 🤔

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -109,19 +112,25 @@ def before_request(
...
             request=request,
             form_data=form_data,
             filename=file.file_name,
         )
-        call_threads = self._get_split_pdf_call_threads()
         self.partition_requests[operation_id] = []
         last_page_content = io.BytesIO()
         last_page_number = 0
-        with ThreadPoolExecutor(max_workers=call_threads) as executor:
+        with ProcessPoolExecutor(max_workers=concurrency_level) as executor:
```
micmarty-deepsense, line 133: @badGarnet here's the change you requested

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -332,67 +345,66 @@ def _decode_content_disposition(self, content_disposition: bytes) -> dict[str, s
         parameters_dict = {p[0]: p[1].strip('"') for p in parameters}
         return parameters_dict
 
+    @staticmethod
```
micmarty-deepsense, line 348: TBH I'm not a fan of static methods, so I if you want me to make all of them a regular functions, feel free to vote here.

Converting this method to static was necessary to pickle arguments correctly (required by ProcessPoolExecutor). 
image

src/unstructured_client/_hooks/custom/split_pdf_hook.py
badGarnet: I would lean toward concurrency level:
- this is technically not a threadpool so this is not controlling number of threads, but number of concurrent processes
- when there are issues debugging threads vs. processes are often different
- as a developer (this variable is exposing to developers) being clear on what the variable controls can help more than reducing different names
- with that said in this client we seem to just have processes so we can stick with concurrency level

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -332,67 +345,66 @@ def _decode_content_disposition(self, content_disposition: bytes) -> dict[str, s
...
         page_content: io.BytesIO,
         filename: str,
         page_number: int,
     ) -> requests.Request:
-        """
-        Creates a request object for a part of a splitted PDF file.
+        """Creates a request object for a part of a splitted PDF file.
 
         Args:
-            request (requests.PreparedRequest): The original request object.
```
micmarty-deepsense, line 384: we have type hints already, so this is unnecessary IMO

awalker4 (approved):
LGTM! Per standup today, we've decided to hold off a bit longer setting the default split to True. We'll flip this when pay per page goes out.

src/unstructured_client/_hooks/custom/split_pdf_hook.py
badGarnet: this is the part that really sets process and thread apart:
- generally a regular function is better as this reinforces the design that the code runs in multiprocessing does not require context more than whats in the function and variables passed in
- multithreading shares the context in the process they spawn from

badGarnet (approved):
there is a caveat that merits revisit: when server returns non 200 sometimes it causes the client to hang (now with process pool it will log errors first so customer could have some cue; thread pool would just silently stuck in GIL hell)

cragwolfe:
> there is a caveat that merits revisit: when server returns non 200 sometimes it causes the client to hang (now with process pool it will log errors first so customer could have some cue; thread pool would just silently stuck in GIL hell)

this is a big one. 1% of requests could easily be non 200

micmarty-deepsense:
> there is a caveat that merits revisit: when server returns non 200 sometimes it causes the client to hang (now with process pool it will log errors first so customer could have some cue; thread pool would just silently stuck in GIL hell)

ohhh😮 

I'll try to reproduce this behavior somehow.

.genignore
micmarty-deepsense: done, ready to resolve

src/unstructured_client/_hooks/custom/split_pdf_hook.py
mpolomdeepsense: Yeah, I was using some vscode extension that adds these automatically when generating doc template, that's why type hints are here (i believe that was me that added documentation here). I have nothing against removing it.

README.md
micmarty-deepsense: description updated

micmarty-deepsense:
> LGTM! Per standup today, we've decided to hold off a bit longer setting the default split to True. We'll flip this when pay per page goes out.

alright, I'm changing the default value to `False`

src/unstructured_client/_hooks/custom/split_pdf_hook.py
micmarty-deepsense: alright, I'm removing them

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -1,43 +1,60 @@
...
-PARTITION_FORM_STARTING_PAGE_NUMBER_KEY = "starting_page_number"
 
 DEFAULT_STARTING_PAGE_NUMBER = 1
+DEFAULT_CONCURRENCY_LEVEL = 5
+MAX_CONCURRENCY_LEVEL = 15
+MIN_PAGES_PER_SPLIT = 2
+MAX_PAGES_PER_SPLIT = 20
+
+
+if platform.system() == "Darwin":
```
micmarty-deepsense, line 44: @badGarnet does this fix the issue you've observed on a mac?

README.md
```diff
@@ -106,18 +106,24 @@ See the [general partition](/docs/models/shared/partitionparameters.md) page for
 
 #### Splitting PDF by pages
 
-In order to speed up processing of long PDF files, set `split_pdf_page=True`. It will cause the PDF
-to be split page-by-page at client side, before sending to API, and combining individual responses
-as single result. This will work only for PDF files, so don't set it for other filetypes.
+In order to speed up processing of long PDF files, `split_pdf_page` can be set to `True` (defaults to `False`). It will cause the PDF to be split at client side, before sending to API, and combining individual responses as single result. This parameter will affect only PDF files, no need to disable it for other filetypes.
```
micmarty-deepsense, line 109: @awalker4 [regarding your comment about PPP](https://github.com/Unstructured-IO/unstructured-python-client/pull/86#pullrequestreview-2061835310), I changed the default value to `False` (also in `overlay_client.yaml`)

badGarnet:
> > there is a caveat that merits revisit: when server returns non 200 sometimes it causes the client to hang (now with process pool it will log errors first so customer could have some cue; thread pool would just silently stuck in GIL hell)
> 
> ohhh😮
> 
> I'll try to reproduce this behavior somehow.



> > there is a caveat that merits revisit: when server returns non 200 sometimes it causes the client to hang (now with process pool it will log errors first so customer could have some cue; thread pool would just silently stuck in GIL hell)
> 
> ohhh😮
> 
> I'll try to reproduce this behavior somehow.

Here is my guess: the retry logic might be causing the issue. We relaunch a process before it closes (or with thread tries to relaunch a thread before current one unlocks the GIL -> thread lock)

src/unstructured_client/_hooks/custom/split_pdf_hook.py
micmarty-deepsense: i guess we can resolve this thread

src/unstructured_client/_hooks/custom/split_pdf_hook.py
micmarty-deepsense: done, refactored

_test_unstructured_client/integration/test_decorators.py
```diff
@@ -1,43 +1,43 @@
...
     [
-        ("_sample_docs/list-item-example-1.pdf", True),  # 1 page
-        ("_sample_docs/layout-parser-paper-fast.pdf", True),  # 2 pages
-        ("_sample_docs/layout-parser-paper.pdf", True),  # 16 pages
-        ("_sample_docs/fake.doc", True),
+        ("_sample_docs/list-item-example-1.pdf", True, "fast"),  # 1 page
+        ("_sample_docs/layout-parser-paper-fast.pdf", True, "fast"),  # 2 pages
+        # NOTE(mike): using "fast" strategy fails on this file for unknown reasons
+        ("_sample_docs/layout-parser-paper.pdf", True, "hi_res"),  # 16 pages
+        ("_sample_docs/fake.doc", True, "fast"),
```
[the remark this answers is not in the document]
micmarty-deepsense: Couldn't find a good, representative PDF example. On https://github.com/Unstructured-IO/unstructured/tree/main/example-docs, we only have tiny PDFs (which won't be helpful when testing the splitting logic), except for this one with 619 pages (https://github.com/Unstructured-IO/unstructured/blob/main/example-docs/DA-619p.pdf), which would slow down CI by about 9 minutes. Are there any better candidates? Please suggest them if so.

badGarnet:
> > > there is a caveat that merits revisit: when server returns non 200 sometimes it causes the client to hang (now with process pool it will log errors first so customer could have some cue; thread pool would just silently stuck in GIL hell)
> > 
> > 
> > ohhh😮
> > I'll try to reproduce this behavior somehow.
> 
> > > there is a caveat that merits revisit: when server returns non 200 sometimes it causes the client to hang (now with process pool it will log errors first so customer could have some cue; thread pool would just silently stuck in GIL hell)
> > 
> > 
> > ohhh😮
> > I'll try to reproduce this behavior somehow.
> 
> Here is my guess: the retry logic might be causing the issue. We relaunch a process before it closes (or with thread tries to relaunch a thread before current one unlocks the GIL -> thread lock)

tested again; with the start method forced to be `fork` this problem goes away on macOS. Tested by mocking http responses from the api.

_test_unstructured_client/integration/test_decorators.py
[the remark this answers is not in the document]
hubert-rutkowski85: Hmm I think the existing layout-parser-paper.pdf with 16 pages is good enough for CI. If the logic works with that much, then it should work with 2x more, 4x more etc. because it's not 1..5 edge case. But what I meant in the comment, was to add other popular filetypes like jpg png txt csv, to test that the non pdf detection works fine and doesn't throw unexpected errors.

README.md
```diff
@@ -106,18 +106,24 @@ See the [general partition](/docs/models/shared/partitionparameters.md) page for
...
 
 Warning: this feature causes the `parent_id` metadata generation in elements to be disabled, as that
 requires having context of multiple pages.
 
-The amount of threads that will be used for sending individual pdf pages, is controlled by
-`UNSTRUCTURED_CLIENT_SPLIT_CALL_THREADS` env var. By default it equals to 5. 
-It can't be more than 15, to avoid too high resource usage and costs.
+The amount of workers utilized for splitting PDFs is dictated by the `split_pdf_concurrency_level` parameter, with a default of 5 and a maximum of 15 to keep resource usage and costs in check.
+The size of each batch of pages (ranging from 2 to 20), is internally determined based on the concurrency level and the total number of pages in the document.
```
hubert-rutkowski85, line 115: But it should be stable for a given document with parameters, right? So next time user can expect to get the same splits? It could be important when there are tables that span multiple pages, then if they're splitted, obviously unstructured will return different output than if they weren't.

hubert-rutkowski85: Btw it could be useful to let users know that ProcessPool is used underneath, so that they know what to expect.

hubert-rutkowski85 (approved):

README.md
micmarty-deepsense:
```suggestion
The amount of workers utilized for splitting PDFs is dictated by the `split_pdf_concurrency_level` parameter, with a default of 5 and a maximum of 15 to keep resource usage and costs in check. The splitting process leverages the `ProcessPoolExecutor` to manage concurrency effectively. 
The size of each batch of pages (ranging from 2 to 20) is internally determined based on the concurrency level and the total number of pages in the document.
```

PR #96, opened 2024-05-29
chore: extend logging

ds-filipknefel:
- Added logs to split page hook to improve transparency on the process.
- Changed log split hook member variable for clarity.
- Added retry counter to retry hook logs.
- Moved logging configuration to `__init__` from the `sdk_init` method of a single hook and removed stdout as target for the logger. Logs now go to stderr (the default).

Tested on  `_sample_docs/layout-parser-paper-fast.pdf` and `_sample_docs/layout-parser-paper.pdf`.
First one will log that it cannot be split if run with pdf split.


src/unstructured_client/_hooks/custom/log_retries_after_error_hook.py
```diff
@@ -11,29 +11,25 @@
...
-            logger.info("Response status code: 500. Sleeping before retry.")
 
-            if bool(response.text):
-                logger.info(response.text)
+        if response is not None and response.status_code // 100 == 5:
+            logger.info("Failed to process a request due to API server error with status code %d."
+                        "Sleeping before retry.",
+                        response.status_code)
+            if response.text:
+                logger.info("Server message - %s", response.text)
```
ds-filipknefel, line 28: `openapi.json` defines that retry should happen on 5XX codes rather then only 500, I've modified logs to match that but I don't know which intention is the desired one.
```json
    "x-speakeasy-retries": {
        "strategy": "backoff",
        "backoff": {
            "initialInterval": 500,
            "maxInterval": 60000,
            "maxElapsedTime": 900000,
            "exponent": 1.5
        },
        "statusCodes": [
            "5xx"
        ],
        "retryConnectionErrors": true
    }
```

src/unstructured_client/_hooks/custom/log_retries_after_error_hook.py
```diff
@@ -1,47 +1,55 @@
 import logging
```
awalker4, line 1: It's probably worth renaming this to logger_hook or something like that.

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -255,19 +278,25 @@ def _await_elements(self, operation_id: str, response: requests.Response) -> Opt
...
             return None
-
-        responses = []
+  
+        responses = [future.result() for future in prt_requests] + [response]
+        successful_responses = []
         elements = []
-        for future in prt_requests:
-            res = future.result()
+        for response_number, res in enumerate(responses):
```
awalker4, line 285: We should 1-index the logs below.

awalker4 (approved):
Looks great!

mpolomdeepsense:
@awalker4 
@ds-filipknefel asked me whether I could finish this task if he couldn't do it before the weekend (he is off today). Applied the changes.

src/unstructured_client/_hooks/custom/split_pdf_hook.py
mpolomdeepsense: Changed

src/unstructured_client/_hooks/custom/log_retries_after_error_hook.py
mpolomdeepsense: Not sure what you mean by this, I saw that default `logging` was used to log some of the messages. Changed it to use custom unstructured client logger.

micmarty-deepsense (approved):

src/unstructured_client/_hooks/custom/log_retries_after_error_hook.py
mpolomdeepsense: Oh, I see now. Thought it was a comment for split_pdf_hook.py. Yep, I agree, should be renamed.

mpolomdeepsense: Changed

PR #102, opened 2024-06-03
chore/enable publishing for python client 🎉 

awalker4:
Changes:
* Bring the publish step back to the speakeasy workflow, regenerate the github action with `speakeasy configure publishing`
* Remove incorrect readme note about parent_id being disabled
* Knock the package version back down to 0.23.0 for continuity with PyPI versions


Coniferish:
Has the client been regenerated since the PR that changed the element types to a list of dicts? When testing on this branch I'm getting element objects that don't have any metadata (regardless of whether the pdf is being split by the client)

awalker4:
Oof, great catch. The new openapi.json hasn't been deployed yet. We'll have to have this pr sit behind that, plus I'll add a unit test to check the output format.

awalker4:
Okay, response schema fixed in #103

badGarnet (approved):

PR #104, opened 2024-06-04
fix/allow asyncio page splitting in nested event loops

awalker4:
We're now using asyncio for page split concurrency, but because the client itself is not async, we need to manage our own event loop. This complains if your environment already has a running event loop. For instance, setting `split_pdf_page=True` in a jupyter cell will give you `RuntimeError: This event loop is already running`.

Turns out there's a simple library to allow for nested event loops. We just apply the monkeypatch in split_pdf_hook.py and the error goes away.

To verify, you'll need to install the current client on main, test in jupyter, then install on this branch and verify the error is gone.
* Checkout the main branch and run `pip install -e .` in your pyenv to install the local version of the client
* Run `make run-jupyter` and open up the sample notebook in `_jupyter/`. 
* Update the sample code to use `split_pdf_page=True`. Change the `server_url` to hit your local server, or add your freemium api key to the snippet. Once the result starts coming back, you should get the above error.
* Checkout this branch and install locally again
* Restart your kernel and try your snippet again. The error should be gone.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ setuptools.setup(
         "jsonpath-python>=1.0.6",
         "marshmallow>=3.19.0",
         "mypy-extensions>=1.0.0",
+        "nest-asyncio>=1.6.0",
         "packaging>=23.1",
         "pypdf>=4.0",
         "python-dateutil>=2.8.2",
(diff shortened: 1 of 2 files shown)


badGarnet (approved):
with a pin that this is only a temporary solution since this dep is an archived repo

PR #105, opened 2024-06-04
fix/add nest-asyncio to gen.yaml

awalker4:
I added it to setup.py which of course is autogenerated.


Coniferish:
Approving, but what is this for?

Coniferish (approved):

awalker4:
I added this as a dep directly in setup.py in #104

PR #109, opened 2024-06-07
Refer to Unstructured docs in readme

awalker4:
Our docs page will be the source of truth, and the readme can have some general information from the autogenerated sections.

Changes:
* Link to our docs at the top
* Remove the manually created usage snippet and swap out for the autogenerated one. This way, the readme stays up to date while we focus on the docs page.
* Bring back autogenerated error handling - this will populate the next time the client regens

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@
   <p>Python SDK for the Unstructured API</p>
 </h2>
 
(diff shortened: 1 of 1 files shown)


Coniferish (approved):
lgtm

PR #111, opened 2024-06-10
fix: remove false success log & improve retry log

ds-filipknefel:
In the PR introducing more logging two things turned out lacking:
- there was a false success log in split-page after error hook
- the retry logs were only for 5XX responses where speakeasy also has retry set for ConnectionError exceptions

This PR removes the faulty success log, moves success/failure logs to the `LoggerHook` and introduces logging for `ConnectionError` retries.


mpolomdeepsense (approved):
LGTM :+1:

awalker4 (approved):

PR #118, opened 2024-06-17
chore/change default split page behavior to true

awalker4:
* Set the split_pdf_page default to true and run `make client-generate` locally.
* Update the readme, add another reference back to our docs
* Change some warning logs to info. The user should not be warned about default behavior for non pdf files

# Testing
Use the client locally and verify that split mode is the default, and that the client behavior is consistent with older versions.

* Set up (or activate) your pyenv for the client: `pyenv virtualenv 3.12 unstructured-client; pyenv activate unstructured-client`
* Check out this branch and install: `pip install -e .`
* Run this sample script in the top level of the client repo. Try different files in `_sample_docs` and verify that the logging and results look acceptable.

```
from unstructured_client.models import shared, operations

import json

api_key = "free-api-key"

s = UnstructuredClient(
    api_key_auth=api_key,
)

    files=shared.Files(
        content=f.read(),
        file_name=filename,
    )

    shared.PartitionParameters(
        files=files,
        strategy=shared.Strategy.AUTO
    ),
)

try:
    print(json.dumps(resp.elements, indent=4))
except Exception as e:
    print(e)
```

### DIFF
--- a/.speakeasy/gen.lock
+++ b/.speakeasy/gen.lock
@@ -1,12 +1,12 @@
 lockVersion: 2.0.0
 id: 8b5fa338-9106-4734-abf0-e30d67044a90
 management:
-  docChecksum: 5365c99c52e23b044ef9916ecf51b1a9
+  docChecksum: c7e23b3b8242eb21eccb2091bcc57c72
   docVersion: 1.0.35
   speakeasyVersion: 1.308.1
   generationVersion: 2.342.6
-  releaseVersion: 0.23.5
-  configChecksum: e210d7bff3afd386269cb7c6adeef630
+  releaseVersion: 0.23.6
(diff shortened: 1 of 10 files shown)


Klaijan (approved):
Tested with non-pdf files, logging shows INFO then print out output elements 
e.g.
```
INFO: Preparing to split document for partition.
INFO: Given file doesn't have '.pdf' extension, so splitting is not enabled.
INFO: Partitioning without split.
INFO: Successfully partitioned the document.
[
    {
        "type": "Title",
        "element_id": "7366bfb62015d7f749e8c38e7284a60c",
        "text": "Lorem ipsum dolor sit amet.",
        "metadata": {
            "category_depth": 0,
            "filename": "fake.doc",
            "languages": [
                "por",
                "cat"
            ],
            "filetype": "application/msword"
        }
    }
]
```

PR #122, opened 2024-06-27
chore: Bump the default split page concurrency

awalker4:
Verified that this shows a speedup by doing a local pip install and running the following snippet before and after the change:

```
from unstructured_client.models import shared

s = UnstructuredClient(
    server_url=SERVER_URL,
    api_key_auth=API_KEY,
    )

filename = "../_sample_docs/layout-parser-paper.pdf"

    # Note that this currently only supports a single file
    files=shared.Files(
        content=f.read(),
        file_name=filename,
	)

    files=files,
    strategy="hi_res",
)

start_time = time.time()
end_time = time.time()
print(f"Elapsed time: {end_time - start_time} seconds")
```


Coniferish (approved):
LGTM once tests pass

awalker4:
Ah, the tests should pass once [this](https://github.com/Unstructured-IO/unstructured-api/pull/436) is merged. The integration tests are pulling the latest `unstructured-api` image and testing requests against it, so we got blocked by the recent docx "None is not a valid mimetype` bug.

PR #125, opened 2024-07-10
feat: Parameter to send custom page range when splitting pdf

awalker4:
# New parameter
Add a client side param called `split_pdf_page_range` which takes a list of two integers, `[start_page, end_page]`. If `split_pdf_page` is `True` and a range is set, slice the doc from `start_page` up to and including `end_page`. Only this page range will be sent to the API. The subset of pages is still split up as needed.

# Other changes
Allow our custom hooks to properly access list parameters, so we're able to intercept `split_pdf_page_range`. We need extra handling to get list params out of the request in `parse_form_data`, and to rebuild the payload in `create_request_body`.

# Testing
Check out this branch and set up a request to your local API:

```
    client = UnstructuredClient(api_key_auth="", server_url="localhost:8000")

        files = shared.Files(
            content=f.read(),
            file_name=filename,
        )

        files=files,
        strategy="fast",
        split_pdf_page=True,
        split_pdf_page_range=[1, 16],
    )

    resp = client.general.partition(req)

```

Test out various page ranges and confirm that the returned elements are within the range. Invalid ranges should throw a ValueError (pages are out of bounds, or end_page < start_page).


src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -154,27 +158,39 @@ def before_request(
...
+            max_pages=len(pdf.pages),
+        )
+
+        page_count = min(len(pdf.pages), page_range_end - page_range_start + 1)
+        logger.info(
+            "Splitting pages %d to %d (%d total)",
+            page_range_start,
+            page_range_end,
+            page_count,
+        )
```
mpolomdeepsense, line 174: Is `min` necesarry here? What I mean is, if the `page_range_end - page_range_start + 1` is higher than `len(pdf.pages)` then the info log will be wrong because the pages range won't match page count.

Also `get_page_range` will throw an error if it's out of bounds so is there a case when `page_count` will be assigned `len(pdf.pages)`?

I might be missing something here so correct me if I'm wrong.

_test_unstructured_client/integration/test_decorators.py
```diff
@@ -110,3 +110,75 @@ def test_integration_split_pdf_for_file_with_no_name():
...
+
+    try:
+        resp = client.general.partition(req)
+    except ValueError as exc:
+        if not expected_ok:
+            assert "is out of bounds." in caplog.text
+            assert "is out of bounds." in str(exc)
+            return
+        else:
+            assert exc is None
```
pawel-kmiecik, line 176: i don't think this works - when an exception object is catched and placed in a variable (`exc` here), it cannot be `None`.

_test_unstructured_client/unit/test_split_pdf_hook.py
```diff
@@ -366,3 +399,38 @@ def test_unit_get_starting_page_number(starting_page_number, expected_result):
...
+            form_data,
+            key=PARTITION_FORM_PAGE_RANGE_KEY,
+            max_pages=20,
+        )
+    except ValueError as exc:
+        if not expected_result:
+            assert "is out of bounds." in str(exc) or "is not a valid page range." in str(exc)
+            return
+        else:
+            assert exc is None
```
pawel-kmiecik, line 434: As above

src/unstructured_client/_hooks/custom/form_utils.py
```diff
@@ -9,14 +9,51 @@
...
+
+    except (ValueError, IndexError) as exc:
+        msg = f"{_page_range} is not a valid page range."
+        logger.error(msg)
+        raise ValueError(msg) from exc
+
+    start, end = page_range
+
+    if not 0 < start <= max_pages or not 0 < end <= max_pages or not start <= end:
+        msg = f"Page range {page_range} is out of bounds. Valid range is (1 - {max_pages})."
```
pawel-kmiecik, line 50: That might be misleading:
```
Valid range is (1 - {max_pages}).
```
Maybe something like:
```
Valid range should be in (1 - {max_pages}).
```

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -154,27 +158,39 @@ def before_request(
...
-        logger.info(
-            "Document split into %d, %d-paged sets.",
-            math.ceil(len(pdf.pages) / split_size),
-            split_size,
-        )
+        pages = pdf_utils.get_pdf_pages(pdf, split_size=split_size, page_start=page_range_start, page_end=page_range_end)
         logger.info(
             "Partitioning %d, %d-paged sets.",
-            math.ceil(len(pdf.pages) / split_size),
+            math.ceil(page_count / split_size),
```
pawel-kmiecik, line 193: I'd introduce a variable like `chunk_no` for this or change the log message to e.g. `Paritioning %d pages chunks, %d-paged sets"

src/unstructured_client/_hooks/custom/split_pdf_hook.py
awalker4: Good catch, we don't need the min since `end - start + 1` will be the correct value. The boundary checks in `get_page_range` ensure that `start` and `end` are less then or equal to `max_pages`, so at most this calculation will be `max_pages - 1 + 1`, or `len(pdf.pages)`.

_test_unstructured_client/integration/test_decorators.py
awalker4: Ah, thanks!

pawel-kmiecik (approved):
LGTM!

PR #128, opened 2024-07-12
Add a make client-generate-local command

awalker4:
For easier testing of local API changes. Run the server at port 5000, and then use `make client-generate-local` to see how the SDK changes.


Coniferish (approved):
LGTM :)

PR #130, opened 2024-07-13
chore: Update usage example to be consistent with docs

awalker4:
The autogenerated example is inconsistent with our hosted docs. Remove this section and copy the usage snippet from the docs. Also, add a note for the new page range feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,41 +35,43 @@ pip install unstructured-client
 ```
 <!-- End SDK Installation [installation] -->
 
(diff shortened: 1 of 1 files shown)


Coniferish (approved):
LGTM.
But I'm also wondering if there's a way for us to continue generating this (and other examples) without them being readily findable by users so we can continue to see how the generated examples change and monitor if there are any significant changes we need to then note in our docs.

PR #134, opened 2024-07-24
Feat/add allow failed param when pdf split is used

pawel-kmiecik:
This PR:
- adds `split_pdf_page_allow_failed` parameter


awalker4 (approved):
LGTM! Don't forget to check in the `overlay_client.yaml` change to make sure the new param stays.

PR #139, opened 2024-07-29
chore: update 401 warning hook for new default url

awalker4:
The default server url is changing to serverless. Therefore, if you get a 401, we should suggest that you meant to use the free api. Also, bump the minor version in anticipation of the url change.


Coniferish (approved):
LGTM

PR #142, opened 2024-07-31
fix: Do not throw error when in a uvloop context

### LINKED ISSUE #133: bug/nest-asyncio conflict with uvloop
**Describe the bug**
The `nest-asyncio` used by `split_pdf_hook.py` triggered the error. The conflict occurs when trying to use `nest-asyncio` together with `uvloop` custom event loop.

> E       RuntimeError: Endpoint http://localhost:6969/api/v1/gen_tables/knowledge/upload_file returned 500 error: {"object":"error","error":"unexpected_error","message":"Opss sorry we ran into an unexpected error. Please try again later.","detail":"Can't patch loop of type <class 'uvloop.Loop'>","request_id":"01906e81-ff60-7966-9f7b-cf3798aecb94","exception":"ValueError"}

awalker4:
PDF page splitting uses asyncio but the SDK is not async. Therefore, we had to manage our own event loop, which can lead to issues in other event loop contexts. Uvloop is one context that does not allow us to use nested event loops. When we find ourselves in a uvloop.Loop, we have to fallback to non splitting mode. #135 will make the whole SDK async so we don't have to hack this.

Closes #133

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ DOCKER_IMAGE ?= downloads.unstructured.io/unstructured-io/unstructured-api:lates
 
 .PHONY: install-test
 install-test:
-	pip install pytest pytest-asyncio pytest-mock requests_mock pypdf deepdiff requests-toolbelt
+	pip install pytest pytest-asyncio pytest-mock requests_mock pypdf deepdiff requests-toolbelt uvloop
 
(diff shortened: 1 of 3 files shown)


pawel-kmiecik (approved):
LGTM!

PR #145, opened 2024-08-06
fix: handle RuntimeError when checking if we can use pdf split mode

awalker4:
Certain environments throw a `RuntimeError: There is no current event loop in thread 'asyncio_0'` when we try to check if we're able to use the splitting code. In this case, return False so we can fallback to non split mode.

Once this is merged, I can regenerate the client to publish 0.25.4.


Coniferish (approved):
Approving, but can we note in a comment in the pr or the code what environments (if we know) this error is raised in?

PR #148, opened 2024-08-08
chore: Remove warning logs from pypdf

awalker4:
When we load a PdfReader with `strict=False`, we can get warnings from pypdf. Unless an error is thrown, these do not affect processing, so let's just suppress the pypdf log level.


MKhalusova (approved):
LGTM! Thanks :)

PR #152, opened 2024-08-12
fix: test retry with 502 status code instead of 500

mackurzawa:
## Description
Status code 500 doesn't retry now, so changed status code from `500` to `502` in `test_unit_retry_with_backoff_does_retry`


pawel-kmiecik (approved):
LGTM

PR #157, opened 2024-08-20
fix: Address mypy typing errors in v2 SDK

awalker4:
Fix all mypy errors due to incorrect typing after the SDK v2 merge (https://github.com/Unstructured-IO/unstructured-python-client/pull/135). Logic changes should be minimal, this is mostly to change type hints where a `httpx.Response` is used instead of a `requests.Response`, etc. I removed some `form_utils.py` functions where we no longer need to convert a httpx request back to Requests. There's more we can cleanup in here, but let's get the V2 migration settled first.

Add mypy to `make lint` so that we can cath these errors before merging. The publish job runs a full linter suite, and these changes made it to main but broke the publish job.

Also, remove the Patch Custom Code step that I added to the generate. This broke the job. There are some minor changes to the Speakeasy code on the main branch. In the short term, this means we'll have to run `make patch-custom-code` whenever we regenerate.

# To verify
Make sure you can lint and run the tests locally. `make lint` and `make test`. You can also verify that the pdf split behavior has not changed with a call to your local server:
```
from unstructured_client.models import shared, operations

import json

s = UnstructuredClient(
    server_url="http://localhost:8000",
)

    files=shared.Files(
        content=f,
        file_name=filename,
    )

        shared.PartitionParameters(
            files=files,
            strategy="fast",
            split_pdf_page_range=[4,8],
        ),
    )

    print(json.dumps(resp.elements, indent=4))
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,8 +13,7 @@ install-test:
 
 .PHONY: install-dev
 install-dev:
-	pip install jupyter uvloop
-	pip install pylint mypy
+	pip install jupyter uvloop pylint mypy
 
 ## install:					installs all test, dev, and experimental requirements
 .PHONY: install
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ install-test:
 
 .PHONY: install-dev
 install-dev:
-	pip install jupyter
+	pip install jupyter uvloop
 	pip install pylint mypy
 
 ## install:					installs all test, dev, and experimental requirements
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ install-test:
 .PHONY: install-dev
 install-dev:
 	pip install jupyter
-	pip install pylint
+	pip install pylint mypy
 
 ## install:					installs all test, dev, and experimental requirements
 .PHONY: install
@@ -48,6 +48,7 @@ test-integration-docker:
 .PHONY: lint
 lint:
 	pylint --rcfile=pylintrc src
+	mypy src
 
 #############
 # Speakeasy #
--- a/src/unstructured_client/_hooks/custom/form_utils.py
+++ b/src/unstructured_client/_hooks/custom/form_utils.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 from typing import Union
 
-from requests_toolbelt.multipart.decoder import MultipartDecoder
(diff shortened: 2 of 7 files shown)


src/unstructured_client/_hooks/custom/form_utils.py
```diff
@@ -35,7 +35,7 @@ def get_page_range(form_data: FormData, key: str, max_pages: int) -> tuple[int,
     try:
         _page_range = form_data.get(key)
 
-        if _page_range is not None:
+        if _page_range is not None and isinstance(_page_range, list):
```
badGarnet, line 38: since `isinstance(None, list) == False` -> this can be shorted to just `isinstance(_page_range, list)`

badGarnet: this applies to the logics below as well; `None` is its own type.

Makefile
```diff
@@ -13,8 +13,8 @@ install-test:
 
 .PHONY: install-dev
 install-dev:
-	pip install jupyter
-	pip install pylint
+	pip install jupyter uvloop
```
pawel-kmiecik, line 16: can we just move it to a single line? or place in some `requirements-dev.txt`?

pawel-kmiecik (approved):
Found a nit, but generally LGTM

pawel-kmiecik (dismissed):
I've just noticed that with V2 poetry is required.
Previously it was just set as a build tool but everything worked with pip.
Now, the dependencies are defined as poetry deps in `pyproject.toml` so they are not installed with the package (like pydantic) if installing with `pip install -e .`.

I think it requires adding some Makefile commands to install the package properly (with `poetry install`) or adding install steps in the README

awalker4:
> I've just noticed that with V2 poetry is required. Previously it was just set as a build tool but everything worked with pip. Now, the dependencies are defined as poetry deps in `pyproject.toml` so they are not installed with the package (like pydantic) if installing with `pip install -e .`.
> 
> I think it requires adding some Makefile commands to install the package properly (with `poetry install`) or adding install steps in the README

Ah, good call! I'm not very familiar with poetry but I suppose this is the time to switch over the makefile commands where it's needed. I'll throw in a new issue for this, as right now I'm trying to unblock the split page retry fixes, and CI does seem to be installing everything as needed.

Coniferish (approved):
LGTM

PR #159, opened 2024-08-22
fix/fix mypy error in generate job

awalker4:
[This job](https://github.com/Unstructured-IO/unstructured-python-client/actions/runs/10501696789/job/29092112690) failed due to a `import-not-found` error from mypy.


amadeusz-ds (approved):
LGTM

PR #160, opened 2024-08-22
fix/ensure that split pdf requests are retried

awalker4:
We discovered that when a pdf is split into smaller chunks, those requests are not being retried. Now that we have `allow_failed=False`, this results in the whole document failing as soon as any of the child requests hit a transient error. The fix is to reuse the `utils.Retry` logic that the main code path uses. Copying the retry config in the hook logic is not great, and we can work with Speakeasy to make the internal logic more modular so we can reuse more. But for now, this will address the current failures while we work on a better implementation.

Testing:
See the added unit test. The existing retry logic works for the final split page, everything else needs to use the new logic. To test this, I mocked a response from the server to return 502 for a low `starting_page_number`, which we know will have to be handled by the hooks.

Other changes:
Remove the "Not splitting" log. When the final split page is retried, it triggers all the hooks again. We need to force `split_pdf_page=False` in this request, and we don't need additional logging when this code is hit again.

### DIFF
--- a/_test_unstructured_client/integration/test_decorators.py
+++ b/_test_unstructured_client/integration/test_decorators.py
@@ -290,8 +290,7 @@ def test_split_pdf_requests_do_retry(monkeypatch):
     """
     Test that when we split a pdf, the split requests will honor retryable errors.
     """
-    # This is in a list so we can pass a reference to the mock function
-    number_of_502s = [1]
+    number_of_502s = 1
 
     async def mock_send(_, request):
         """
@@ -306,9 +305,10 @@ def test_split_pdf_requests_do_retry(monkeypatch):
         decoded_body = MultipartDecoder(request_body, request.headers.get("Content-Type"))
         form_data = form_utils.parse_form_data(decoded_body)
 
-        if number_of_502s[0] > 0:
+        nonlocal number_of_502s
+        if number_of_502s > 0:
             if "starting_page_number" in form_data and int(form_data["starting_page_number"]) < 3:
-                number_of_502s[0] -= 1
+                number_of_502s -= 1
                 return Response(502, request=request)
 
         mock_return_data = [{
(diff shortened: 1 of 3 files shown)


_test_unstructured_client/integration/test_decorators.py
```diff
@@ -275,3 +284,69 @@ def test_integration_split_pdf_strict_mode(
...
     )
     assert len(diff) == 0
+
+
+def test_split_pdf_requests_do_retry(monkeypatch):
+    """
+    Test that when we split a pdf, the split requests will honor retryable errors.
+    """
+    # This is in a list so we can pass a reference to the mock function
+    number_of_502s = [1]
```
pawel-kmiecik, line 294: nit: using `nonlocal` in `mock_send` instead of list would be more elegant and clear ;)
https://www.w3schools.com/python/ref_keyword_nonlocal.asp

src/unstructured_client/_hooks/custom/request_utils.py
```diff
@@ -69,10 +70,34 @@ async def call_api_async(
...
 
 
+async def send_request_async_with_retries(client: httpx.AsyncClient, request: httpx.Request):
+    retry_config = utils.RetryConfig(
+        "backoff",
+        utils.BackoffStrategy(2000, 60000, 1.5, 900000),
+        True
+    )
+
+    retryable_codes = [
```
pawel-kmiecik, line 84: I guess it's a hard to eliminate tradeoff to hardcode here - the SDK uses `openapi.json` to read these values.

pawel-kmiecik (approved):
Found a nit but LGTM!

_test_unstructured_client/integration/test_decorators.py
awalker4: Perfect, thanks!

src/unstructured_client/_hooks/custom/request_utils.py
awalker4: Yes, I would love to be able to create the new httpx requests and just send them back into `general.partition`, to reuse the retry config and any other client settings. We'll need speakeasy to add a reference to the sdk in the hook context, but then we'll be able to really clean up `split_pdf_hook.py`. 

I added a comment to make it clear why this is hardcoded.

PR #163, opened 2024-08-24
Update request_utils.py

cragwolfe:
Tweak retry values to be more robust:

Should be equivalent to wait-time-values-between-retries and cumulative-time-between-tries of (in seconds):

```
>>> a = [min(math.floor(3 * (1.88 ** i)), 720) for i in range(10)]; a; [sum(a[:i+1]) for i in range(len(a))]
[3, 5, 10, 19, 37, 70, 132, 249, 468, 720]
[3, 8, 18, 37, 74, 144, 276, 525, 993, 1713]
```


awalker4 (approved):

awalker4:
LGTM! If this works well, we can persist in the api spec which will update the values [over here](https://github.com/Unstructured-IO/unstructured-python-client/blob/5ae2f46a723007259b0369732fa3def6cc53f5f2/src/unstructured_client/general.py#L62).

PR #164, opened 2024-08-27
chore: Set 0.26.0 version for release

awalker4:
Set the version in gen.yaml, to be picked up in the next generate/publish workflow. And fix an import that was flagged by pylint's `consider-using-from-import`

This is in order to release the split pdf retry fix. This will also push out the new `partition_async`, but we can follow up with docs/announcements later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,6 @@
   <p>Python SDK for the Unstructured API</p>
 </h2>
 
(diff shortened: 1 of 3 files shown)


MichaelMcCulloch-deepsense-ai (approved):

awalker4:
Closing - we aren't ready to release yet.

PR #170, opened 2024-09-04
Fix/retry fixes 0.26.0

awalker4:


JOSHMT0744 (approved):
Really like these changes :)
On test_decorators.py, if endpoint url isn't "/general/v0/general", this could cause issues..?

awalker4:
> Really like these changes :) On test_decorators.py, if endpoint url isn't "/general/v0/general", this could cause issues..?

Thanks for the review! For that test - I was able to change the logic that required that check, so I can remove that now.

src/unstructured_client/_hooks/custom/request_utils.py
```diff
@@ -50,6 +52,51 @@ def create_request_body(
     return body
 
 
+async def retry_with_backoff_async(
```
jordan-homan, line 55: is it not possible to use the existing `RetryConfig` class defined in `retries.py` instead of creating a new one?

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -250,8 +273,17 @@ def before_request(
                 page_count % split_size,
             )
 
+        # Use a variable to adjust the httpx client timeout, or default to 30 minutes
+        # When we're able to reuse the SDK to make these calls, we can remove this var
+        # The SDK timeout will be controlled by parameter
+        client_timeout_minutes = 30
+        if timeout_var := os.getenv("UNSTRUCTURED_CLIENT_TIMEOUT_MINUTES"):
```
jordan-homan, line 280: does the client user have to know to set this environment variable?  if so should the default go the other way around maybe? just thinking about how it might be challenging waiting up to 30 minutes without realizing why it's hanging so long

src/unstructured_client/_hooks/custom/request_utils.py
awalker4: Yes, let me switch that back. I copied it over because there was a bug with the autogenerated function (it was using a blocking `time.sleep`, and it didn't have any debug logging. I believe these are both fixed in the latest speakeasy update.

src/unstructured_client/_hooks/custom/split_pdf_hook.py
awalker4: We've had some users get stuck here. For very large pdfs, a hi_res call can sit for 30 minutes, at which point they get a httpx.ReadTimeout, the doc gets retried, and the same thing happens again. This should be rare, but we wanted the timeout to account for the worst case. And we added the env var so we could ask users to bump the timeout manually if this still wasn't resolved.

There is a http timeout parameter for the SDK now, but due to some gaps in the hook context, we don't have access to that param from down here. Longer term, that should be the value that users can configure. The env var here is a bit of a bandaid, so I'm keeping it out of the readme, etc.

PR #175, opened 2024-09-16
fix: Allow split page logic to process files concurrently

awalker4:
# The Issue

`split_pdf_hook.py` does not support multiple concurrent files. This is because we store the split request tasks in `self.coroutines_to_execute[operation_id]`, where `operation_id` is just the string "partition". Therefore, if we send two concurrent docs using the same SDK, they'll both try to await the same list of coroutines. This could result in interleaved results, but mostly it breaks with `RuntimeError: coroutine is being awaited already`, as the second request gets ready to await its requests. This will block anyone trying to use the new `partition_async` to fan out their pdfs.

Note that the js/ts client also has this issue.

# The fix

We need to use an actual id to index into `coroutines_to_execute`. In `before_request`, let's make a uuid and build up the list of coroutines for this doc. We need to pass this id to `after_success` in order to retrieve the results, so we can set it as a header on our "dummy" request that's returned to the SDK.

# Testing

See the new integration test. We can verify this by sending two docs serially, and then with `asyncio.gather`, and confirm that the results are the same.

### DIFF
--- a/src/unstructured_client/_hooks/custom/split_pdf_hook.py
+++ b/src/unstructured_client/_hooks/custom/split_pdf_hook.py
@@ -5,6 +5,7 @@ import io
 import logging
 import os
 import math
+import uuid
 from collections.abc import Awaitable
 from typing import Any, Coroutine, Optional, Tuple, Union, cast
 
@@ -198,7 +199,12 @@ class SplitPdfHook(SDKInitHook, BeforeRequestHook, AfterSuccessHook, AfterErrorH
         # This allows us to use an event loop in an env with an existing loop
         # Temporary fix until we can improve the async splitting behavior
         nest_asyncio.apply()
-        operation_id = hook_ctx.operation_id
+
+        # This is our key into coroutines_to_execute
+        # We need to pass it on to after_success so
+        # we know which results are ours
+        operation_id = str(uuid.uuid4())
+
         content_type = request.headers.get("Content-Type")
 
         request_content = request.read()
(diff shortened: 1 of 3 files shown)


src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -329,14 +335,20 @@ async def call_api_partial(page):
...
         # dummy_request = httpx.Request("GET",  "http://no-op")
-
-        dummy_request = httpx.Request("GET",  "https://api.unstructuredapp.io/general/docs")
+        dummy_request = httpx.Request(
+            "GET",
+            "https://api.unstructuredapp.io/general/docs",
+            headers={"request_id": operation_id},
+        )
 
         return dummy_request
```
badGarnet, line 353: could just 
```suggestion
        return httpx.Request(
            "GET",
            "https://api.unstructuredapp.io/general/docs",
            headers={"request_id": operation_id},
        )
```

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -329,14 +335,20 @@ async def call_api_partial(page):
...
         # Note: We need access to the async_client from the sdk_init hook in order to set
         # up a mock request like this.
         # For now, just make an extra request against our api, which should return 200.
         # dummy_request = httpx.Request("GET",  "http://no-op")
-
-        dummy_request = httpx.Request("GET",  "https://api.unstructuredapp.io/general/docs")
+        dummy_request = httpx.Request(
+            "GET",
+            "https://api.unstructuredapp.io/general/docs",
+            headers={"request_id": operation_id},
```
badGarnet, line 350: should we call it `operation_id` in the header key as well to avoid confusion?

awalker4: Good call!

badGarnet (approved):

PR #181, opened 2024-09-20
chore: revert custom changes to autogenerated code

awalker4:
# The issue
The v2 python client has some minor breaking changes for our users, both related to keyword parameters:

* The PartitionRequest class is now a pydantic model which does not support positional arguments in the constructor. Therefore we always have to name the `partition_parameters` field.

```
        shared.PartitionParameters(
            ....
            ....
        )
    )
```

becomes

```
        partition_parameters=shared.PartitionParameters(
            ....
            ....
        )
    )

```

* Likewise, the function signature for `general.partition` no longer allows for positional args.

Instead of 

```
    response = client.general.partition(req)
```

we have to do 

```
    response = client.general.partition(request=req)
```

# The first approach

These are minimal changes, but they will disrupt a number of users who have copied our sample snippets in the past. My workaround for this was to manually patch the autogenerated code to bring back positional params. However, this will be messy in the long term, as we now have to apply our changes after every regenerate. Instead, let's embrace the version bump to 0.16.0 and just document the breaking change.

Update: The python snippet in the serverless dashboard uses keyword args, so users who copied this out are unaffected.

# Changes

- Revert my manual edits to:
  - Allow `request` as a positional arg in `general.partition` (`general.py`)
  - Add a constructor to `PartitionRequest` so we don't have to use a keyword arg
- Remove the patch file containing these changes
- Update some makefile commands to reuse in other repos


Coniferish (approved):
LGTM

PR #184, opened 2024-09-25
chore Use pdf library to check file without extension

yuming-long:
### Summary

Instead of manually checking filename with `.pdf` extension and return is_pdf = false -> use currently pdf library to read file content to decide if the file content is valid pdf

### DIFF
--- a/src/unstructured_client/_hooks/custom/pdf_utils.py
+++ b/src/unstructured_client/_hooks/custom/pdf_utils.py
@@ -69,7 +69,6 @@ def is_pdf(file: shared.Files) -> bool:
         content = cast(bytes, file.content)
         PdfReader(io.BytesIO(content), strict=True)
     except (PdfReadError, UnicodeDecodeError):
-        logger.info("Loading PDF failed, so splitting is not enabled.")
         return False
 
     return True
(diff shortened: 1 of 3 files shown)


awalker4:
LGTM! I'd just suggest removing the logging when a file can't be loaded in pypdf. For a `.doc`, I see this:

```
INFO: Preparing to split document for partition.
ERROR: 'utf-8' codec can't decode byte 0xd0 in position 0: invalid continuation byte
WARNING: The file does not appear to be a valid PDF.
INFO: Partitioning without split.
```

If any non pdf is expected to hit this path, we can remove the scary ERROR and WARNING.

src/unstructured_client/_hooks/custom/pdf_utils.py
```diff
@@ -56,25 +56,20 @@ def get_pdf_pages(
...
-        return False
 
     try:
         content = cast(bytes, file.content)
         PdfReader(io.BytesIO(content), strict=True)
-    except (PdfReadError, UnicodeDecodeError) as exc:
-        logger.error(exc)
-        logger.warning("The file does not appear to be a valid PDF.")
+    except (PdfReadError, UnicodeDecodeError):
+        logger.info("Loading PDF failed, so splitting is not enabled.")
```
awalker4, line 72: Sorry to be nitpicky, but I would skip logging anything in this path. Since the user will see the word failed and think something's wrong with their non PDF file. Plus we already have some noisy logs to clean up

awalker4 (approved):
LGTM pending the last comment!

PR #188, opened 2024-10-04
chore: Update README and cut 0.26.0 for publishing

awalker4:
Bring back some of the autogenerated README content and make sure our manual sections are using the right syntax. Once we merge and regenerate, 0.26.0 will be published.


Coniferish (approved):
LGTM

PR #192, opened 2024-10-11
fix: Use configured server_url for split page dummy request

### LINKED ISSUE #191: bug/hardcoded GET request to api.unstructuredapp.io causes timeout behind firewall
**Describe the bug**
There is a hardcoded dummy GET request to api.unstructuredapp.io.  There does not appear to be an easy way to override this, and it also doesn't look necessary. This causes a timeout error behind a firewall and makes the unstructured-client unusable in this environment.  Could this just use the "server_url" parameter for unstructured client? 

https://github.com/Unstructured-IO/unstructured-python-client/blob/18a13093686bca1f97cc061f667fec7db21c489c/src/unstructured_client/_hooks/custom/split_pdf_hook.py#L349

**To Reproduce**
Use the client with personal unstructured-api deployment that is behind a firewall.
```

s= UnstructuredClient(,
        server_url=my_server_url,
        api_key_auth=None
    )
res = s.general.partition(request=req)

```
**Expected behavior**
Request hangs on GET request and will eventually timeout. 

**Environment Info**
unstructured-client=0.26.0

awalker4:
Fixes #191. Due to a limitation in our templated codebase, we need our split page logic to return a request that's just going to give a 200 response. Then, we jump back into the AfterSuccessHook and await all of the pdf splits. Hopefully we can clean this up soon to avoid having an extra call at all, but for now let's make sure the GET /general/docs is done on the correct base url.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 0.26.1
+
+### Enhancements
+
+### Features
+
+### Fixes
+* Use the configured server_url for our split page "dummy" request
+
 ## 0.26.0
 
 ### Enhancements
--- a/_test_unstructured_client/integration/test_decorators.py
+++ b/_test_unstructured_client/integration/test_decorators.py
@@ -304,12 +304,14 @@ async def test_split_pdf_requests_do_retry(monkeypatch):
 
         We want to make sure both code paths are retried.
         """
-        # Assert that the SDK issues our no-op request
+        # Assert that the SDK issues our dummy request
         # returned by the BeforeRequestHook
         nonlocal mock_endpoint_called
-        if request.url.host == "no-op" or "docs" in request.url.path:
+        if request.url.host == "localhost" and "docs" in request.url.path:
             mock_endpoint_called = True
             return Response(200, request=request)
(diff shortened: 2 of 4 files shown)


Coniferish (approved):
LGTM.

Just noting this part in `src/unstructured_client/sdk.py` that this change relates to:
```
        current_server_url, *_ = self.sdk_configuration.get_server_details()
        server_url, self.sdk_configuration.client = hooks.sdk_init(
            current_server_url, self.sdk_configuration.client
        )
        if current_server_url != server_url:
            self.sdk_configuration.server_url = server_url
```

awalker4:
> LGTM.
> 
> Just noting this part in `src/unstructured_client/sdk.py` that this change relates to:
> 
> ```
>         current_server_url, *_ = self.sdk_configuration.get_server_details()
>         server_url, self.sdk_configuration.client = hooks.sdk_init(
>             current_server_url, self.sdk_configuration.client
>         )
>         if current_server_url != server_url:
>             self.sdk_configuration.server_url = server_url
> ```

Ah good catch! This hook will always have to run first - let me get a unit test in for that.

PR #202, opened 2024-10-30
feat: improved memory management during long split pdf processing

pawel-kmiecik:
This PR:
- improves the memory management by freeing the memory early when processing larger files, thus holding the data in memory for a long time
- instead, it uses tempfiles to keep:
  - pdf_chunks for split_pdf_page functionality
  - partial response elements as json files


src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -270,7 +285,27 @@ def before_request(
...
+        pages = self._get_pdf_pages(
+            pdf,
+            operation_id=operation_id,
+            split_size=split_size,
+            page_start=page_range_start,
+            page_end=page_range_end
+        )
+        # force free PDF object memory
+        del pdf
+        logger.info(
```
pawel-kmiecik, line 297: ahh must have left these while rebasing with main - will be removed in next commit

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -288,22 +311,36 @@ async def call_api_partial(page):
...
                 )
 
+                # Immediately delete request to save memory
+                del response._request  # pylint: disable=protected-access
+                response._request = None  # pylint: disable=protected-access
+
+                # If we get 200, dump the contents to a file and return the path
+                temp_dir = self.tempdirs[operation_id]
+                if response.status_code == 200:
+                    temp_file_name = f"{temp_dir.name}/{uuid.uuid4()}.json"
```
jordan-homan, line 325: Is there a risk that the customer using the client doesn't have permissions to write files in the current process they're running?

pawel-kmiecik: Good point. But at the same time `unstructured` OS uses temp files heavily. Maybe making the temp dir parametrized via env could resolve this?

jordan-homan: yeah would guess for our own environments it's safe but a customer could hit an error. maybe we could fall back to using memory if we don't have write access (not sure on implementation there), or we could expose a path in the client that the caller sets up explicitly for temporary storage maybe?

pawel-kmiecik:
FYI: Added more optimizations and on 1k pdf file the memory usage dropped from ~250MB to ~120MB (<100MB if not reading the contents of a file).
Next steps include feature flag for memory optimizations using tmp files caching (as @jordan-homan mentioned - it might be a problem for some users).

src/unstructured_client/_hooks/custom/split_pdf_hook.py
pawel-kmiecik: I've added a feature flag in partition parameters (feat disabled by default)

src/unstructured_client/models/shared/partition_parameters.py
```diff
@@ -124,6 +124,10 @@ class PartitionParametersTypedDict(TypedDict):
     r"""This parameter determines if the PDF file should be split on the client side. It's an internal parameter for the Python client and is not sent to the backend."""
     split_pdf_page_range: NotRequired[List[int]]
     r"""When `split_pdf_page is set to `True`, this parameter selects a subset of the pdf to send to the API. The parameter is a list of 2 integers within the range [1, length_of_pdf]. A ValueError is thrown if the given range is invalid. It's an internal parameter for the Python client and is not sent to the backend."""
+    split_pdf_cache_tmp_data: NotRequired[bool]
```
jordan-homan, line 127: we can't edit these files manually as they will be removed the next time we sync from speakeasy

pawel-kmiecik: I must have mixed up the files where we define new custom params - on it.

jordan-homan:
before merging we should squash into one commit so that it's easy to follow in the public git history

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -439,3 +665,6 @@ def _clear_operation(self, operation_id: str) -> None:
         """
         self.coroutines_to_execute.pop(operation_id, None)
         self.api_successful_responses.pop(operation_id, None)
+        tempdir = self.tempdirs.pop(operation_id, None)
+        if tempdir:
```
jordan-homan, line 669: are there other areas we should consider cleaning up or does `_clear_operation` cover every case? curious if there are scenarios where if the api call fails, we don't clean up and leave files in the directory

pawel-kmiecik:
> before merging we should squash into one commit so that it's easy to follow in the public git history

Isn't it the default setting? I'm not sure what's the squash commit message setting in this repo (It should be the PR description I guess)
Screenshot 2024-11-07 at 15 28 45

src/unstructured_client/_hooks/custom/split_pdf_hook.py
```diff
@@ -270,40 +334,47 @@ def before_request(
...
+            pdf_chunk_paths = self._get_pdf_chunk_paths(
+                pdf,
+                operation_id=operation_id,
+                split_size=split_size,
+                page_start=page_range_start,
+                page_end=page_range_end
+            )
+            # force free PDF object memory
+            del pdf
+            pdf_chunks = self._get_pdf_chunk_files(pdf_chunk_paths)
```
jordan-homan, line 347: Question in general about the new process: how do we improve memory management in general?

So for example, if I have a 5000 page PDF that consumes ~50MB of memory per chunk (let's say this process breaks it down into 100 batches of 50 pages), I understand that when batch 1 of 50mb, it will write that to a file, batch 2 write to file and so on...


so if the job takes 20 minutes to complete, the overall memory usage stays low while we write to files.

However, at the end of a 20 minute time-frame once the last (50th) batch finally completes, we then take all of the 100 batches written in each file, and bring everything back into memory at this point in time to return it to the caller, right?  Are we optimizing purely for the duration of the run to keep memory low, but not concerned that, at the end the memory needs to go back up to resolve the file?

jordan-homan:
> Isn't it the default setting? I'm not sure what's the squash commit message setting in this repo (It should be the PR >
 sounds good!

src/unstructured_client/_hooks/custom/split_pdf_hook.py
pawel-kmiecik: Yes, you're right - we don't want to occupy the memory in a long process (and with hi res it can be tens or hundreds of minutes). But we cannot avoid loading the json files back to the memory because the interface must work the same way as for non `split-pdf` calls - so returning a Response with a json attached. That's why we won't avoid the spike at the end.

 Maybe we could replace the response `read` operation with something smart that would load the files one by one but I don't think we'll get much here.

Also - what's important in these changes - we no longer pass the bytes or io.BytesIO object to the request but a file object which is read to the socket in a buffered way.

src/unstructured_client/_hooks/custom/split_pdf_hook.py
pawel-kmiecik: That's the limitation of the SDK... it's not supposed to fail cleaning on the condition that all the possible exceptions are caught (and it looks we catch them - but it's not a fireproof :D )
It would be much cleaner and resilient if we could use a nice context manager but because of the hooks we hack here we cannot.

jordan-homan (approved):

PR #204, opened 2024-10-31
increase default timeout to 10 minutes

vangheem:


src/unstructured_client/sdk.py
```diff
@@ -28,7 +28,7 @@ def __init__(
         client: Optional[HttpClient] = None,
         async_client: Optional[AsyncHttpClient] = None,
         retry_config: OptionalNullable[RetryConfig] = UNSET,
-        timeout_ms: Optional[int] = None,
+        timeout_ms: int = 10 * 60 * 1000,
```
jordan-homan, line 31: can't modify these files it's auto generated

PR #212, opened 2024-11-26
add test for chunking strategies

Coniferish:


qued:
Thanks @Klaijan for investigating this. @Coniferish is the expected behavior that each page becomes a single chunk? This isn't what I get from the documentation. As far as we can tell the chunking behavior is behaving correctly, but I'm open to exploring further if someone can elaborate on what it's doing that's wrong.

PR #226, opened 2025-02-18
fix/Address server_url breaking change

awalker4:
In order to add the platform-api functions, we now need to handle two openapi specs. When these are merged, we get a breaking change with the `server_url` arg:

```
# The passed url is ignored and we go to our default Serverless URL. 
# You suddenly get an invalid api key error
s = UnstructuredClient(server_url="my_own_url")
s.general.partition()

#  This does the right thing
s = UnstructuredClient()
s.general.partition(server_url="my_own_url")
```

We published the 0.30.0 client with this change, but let's do the right thing before it breaks a whole lot of users. For now, let's directly edit the endpoint functions and make sure the client level url is still accepted. We can work with Speakeasy to roll this back into the auto generated SDK.

## Changes
- Add a helper function to our `clean_server_url` hook file to select the right url from three options: the endpoint, the client, or the default from the config
- Call this helper from every generated method (plus async versions)
- Add a `test_server_urls.py` to test every URL case against every endpoint. The user can pass the URL in either location, and the SDK will strip the path if it's given, for backwards compat.
- Add the endpoint files to `.genignore`. This will unblock the generate action for now, and our test suite will tell us if these files have gotten stale.

Other:
- Update `workflow.yaml` to reorder the openapi specs. Both specs include `HttpValidationError`, and the later spec will overwrite the duplicate. This schema changed a bit in Serverless, so some tests are failing on main. Note that the `make client-generate-sdk` does the right thing, but this got reverted when the github generate action ran again.

## Testing
- Check out this branch and verify that your old SDK code with a custom URL is still working


src/unstructured_client/destinations.py
```diff
@@ -34,12 +35,15 @@ def create_destination(
         if timeout_ms is None:
             timeout_ms = self.sdk_configuration.timeout_ms
 
-        if server_url is not None:
-            base_url = server_url
-        else:
-            base_url = operations.CREATE_DESTINATION_SERVERS[
+        client_url, *_ = self.sdk_configuration.get_server_details()
+
+        base_url = choose_server_url(
```
pawel-kmiecik, line 40: I'm not sure if I get why do we need to overwrite the platform-specific operations? As I understand that it didn't break anything yet and users can only use our prod environment, hardcoded in each operation.
The only thing we want to overwrite is `partition` operation which got broken.

I'd limit the changes only to that operation - it's rather "frozen" while we'll be working on platform-specific operations and might need regenerating these files.

src/unstructured_client/_hooks/custom/clean_server_url_hook.py
```diff
@@ -7,31 +7,61 @@
...
 
-class CleanServerUrlSDKInitHook(SDKInitHook):
-    """Hook fixing common mistakes by users in defining `server_url` in the unstructured-client"""
+def clean_server_url(base_url: str) -> str:
+    """Fix url scheme and remove the '/general/v0/general' path."""
+
+    if not base_url:
+        return ""
+    # -- add a url scheme if not present (urllib.parse does not work reliably without it)
+    if "http" not in base_url:
```
PastelStorm, line 16: I would parse the URL first and then check if a url has a scheme by using `parsed_url.scheme`. Then we can prepend `http://` or `https://`

src/unstructured_client/_hooks/custom/clean_server_url_hook.py
```diff
@@ -7,31 +7,61 @@
...
+        base_url = "http://" + base_url
+
+    parsed_url: ParseResult = urlparse(base_url)
+
+    if "api.unstructuredapp.io" in base_url:
```
PastelStorm, line 21: I'd use `parsed_url.netloc` here.

src/unstructured_client/_hooks/custom/clean_server_url_hook.py
```diff
@@ -7,31 +7,61 @@
...
+    Helper function to fix a breaking change in the SDK past 0.30.0.
+    When we merged the partition and platform specs, the client server_url stopped working,
+    and users need to pass it in the endpoint function.
+    For now, call this helper in the generated code to set the correct url.
+    """
 
-        parsed_url: ParseResult = urlparse(base_url)
+    # First, see if the endpoint has a url:
+    # s.general.partition(server_url=...)
+    if endpoint_url is not None:
```
PastelStorm, line 39: This means the endpoint_url is treated as an optional kwarg and it's currently not.

src/unstructured_client/_hooks/custom/clean_server_url_hook.py
```diff
@@ -7,31 +7,61 @@
...
 
-    def clean_server_url(self, base_url: str) -> str:
-        """Fix url scheme and remove the '/general/v0/general' path."""
 
-        if not base_url:
-            return ""
-        # -- add a url scheme if not present (urllib.parse does not work reliably without it)
-        if "http" not in base_url:
-            base_url = "http://" + base_url
+def choose_server_url(endpoint_url, client_url, default_endpoint_url) -> str:
```
PastelStorm, line 29: I feel like this entire helper function can be written as a oneliner:
`url = endpoint_url if endpoint_url is not None else (client_url or default_endpoint_url)`
the logic and reasoning behind this can be explained in the docstring.

src/unstructured_client/_hooks/custom/clean_server_url_hook.py
awalker4: That comment has been around for a while - I tried parsing without the scheme and it does look like we need to make sure `http://` is there first.

```
In [2]: urlparse("localhost:8000/general/v0/general")
Out[2]: ParseResult(scheme='localhost', netloc='', path='8000/general/v0/general', params='', query='', fragment='')
```

src/unstructured_client/_hooks/custom/clean_server_url_hook.py
awalker4: This can come in from the generated code as None over [here](https://github.com/Unstructured-IO/unstructured-python-client/blob/a4bfd93eefbd909e1287f3d066f156b365b4a2e4/src/unstructured_client/general.py#L41)

src/unstructured_client/_hooks/custom/clean_server_url_hook.py
PastelStorm: Oh I see, good to know!

PastelStorm (approved):

PR #243, opened 2025-03-04
fix: Fix job status in contract tests

awalker4:
We have a JobStatus schema as of https://github.com/Unstructured-IO/unstructured-python-client/pull/242

We need to change the mocked "active" to one of the valid job statuses.


pawel-kmiecik (approved):
LGTM!

PR #260, opened 2025-04-09
fix cleaning urls

vangheem:
Right now, if an API is hosted at a subpath, this will strip out the path.


tabossert (approved):

awalker4:
@yuming-long has a ticket to fix this, it involves patching some of the autogenerated code for now. Closing this one.

PR #267, opened 2025-04-22
feat: fix CI failing issues

yuming-long:
fix issues on main CI:
* test_contract (3.9) stuck 6hrs... should really be just 30s
* test_integration (3.9) failling but not on other version

hopefully can speed up CI a bit with `pytest -n auto` (Automatically run your tests in parallel across all available CPU cores) -> ubuntu-latest has 2 core CPU so should be able to speed up in half?

note:
* maybe could be faster if we switch to our own github runner? (if it has more core)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -79,7 +79,7 @@ jobs:
       - name: Install dependencies
         run: |
           make install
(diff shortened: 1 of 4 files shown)


amanda103 (approved):
👍🏻

PR #273, opened 2025-05-28
Remove contract tests that do not test the python client and fix depe…

jiajun-unstructured:
- remove the tests that tests the unstructured library instead of the python client
- remove unstructured from the project's dependency list


awalker4 (approved):
Awesome!!

PR #277, opened 2025-06-18
fix: Return error for invalid PDFs

mpolomdeepsense:
- Adds function for checking PDF files before the request gets sent. Throws appropriate error message in case the file is invalid.


src/unstructured_client/_hooks/custom/pdf_utils.py
```diff
@@ -33,3 +42,34 @@ def read_pdf(pdf_file: Union[BinaryIO, bytes]) -> Optional[PdfReader]:
...
+    """
+    try:
+        # This will raise if the file is encrypted
+        pdf.metadata  # pylint: disable=pointless-statement
+
+        # This will raise if the file's root object is corrupted
+        pdf.root_object  # pylint: disable=pointless-statement
+
+        # This will raise if the file's pages are corrupted
+        list(pdf.pages)
```
pawel-kmiecik, line 65: Did we profiled it for memory/time for a large pdfs (like ~1k pages)?

pawel-kmiecik (approved):
Generally looks ok - I'm just worried a bit about the check's influence on the performance (but I think we do these operations in split-pdf function anyway).

src/unstructured_client/_hooks/custom/pdf_utils.py
mpolomdeepsense: I did not, but maybe we could just check for the first page? I think it will result with the same error. WDYT?

mpolomdeepsense: Tested. This does not slow down execution and doesn't affect memory usage, even with big files (tested on 200, 1000 and 10000 pages pdfs).

Results for 10k pages pdf:
[mem_profiler_results.txt](https://github.com/user-attachments/files/20796733/mem_profiler_results.txt)

PR #279, opened 2025-06-27
feat: add a encrypt_secret helper function

awalker4:
Add a `users.encrypt_secret` that conforms to Unstructured's secret formats. This will allow users to encrypt credentials locally before using them in connector config.

### DIFF
--- a/_test_unstructured_client/unit/test_encryption.py
+++ b/_test_unstructured_client/unit/test_encryption.py
@@ -1,0 +1,128 @@
+from cryptography import x509
+from cryptography.hazmat.primitives import serialization, hashes
+from cryptography.hazmat.primitives.asymmetric import padding, rsa
+from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+from cryptography.hazmat.backends import default_backend
+import os
+import base64
+from typing import Optional
+
+import pytest
+
+from unstructured_client import UnstructuredClient
+
+@pytest.fixture
+def rsa_key_pair():
+    private_key = rsa.generate_private_key(
+        public_exponent=65537,
+        key_size=2048,
+        backend=default_backend()
+    )
+    public_key = private_key.public_key()
+
+    private_key_pem = private_key.private_bytes(
+        encoding=serialization.Encoding.PEM,
+        format=serialization.PrivateFormat.TraditionalOpenSSL,
+        encryption_algorithm=serialization.NoEncryption()
+    ).decode('utf-8')
+
+    public_key_pem = public_key.public_bytes(
+        encoding=serialization.Encoding.PEM,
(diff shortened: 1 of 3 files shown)


src/unstructured_client/users.py
```diff
@@ -458,3 +467,125 @@ async def store_secret_async(
...
+            ).decode('utf-8')
+
+        else:
+            public_key_pem = encryption_cert_or_key_pem
+
+        # If the plaintext is short, use RSA directly
+        # Otherwise, use a RSA_AES envelope hybrid
+        # The length of the public key is a good hueristic
+        if not encryption_type:
+            encryption_type = "rsa" if len(plaintext) <= len(public_key_pem) else "rsa_aes"
```
CyMule, line 585: Can we load the key in this function and then check the size like
```python
    max_payload = (public_key.key_size // 8) - 42  # OAEP overhead
    plaintext_bytes = len(plaintext.encode('utf-8'))
    encryption_type = "rsa" if plaintext_bytes <= max_payload else "rsa_aes"
```

I think this would fail when in some cases. For example
1. Using RSA 2048. Max Payload is ((2048/8 bits) - 42 byte OAEP) = 214 bytes
2. 215 byte plaintext. 215 < 451 PEM which means RSA is selected
3. Attempt to encrypt 215 byte plain text but max payload is 214

src/unstructured_client/users.py
```diff
@@ -458,3 +467,160 @@ async def store_secret_async(
...
+
+        if not isinstance(public_key, rsa.RSAPublicKey):
+            raise TypeError("Public key must be a RSA public key for encryption.")
+
+        # If the plaintext is short, use RSA directly
+        # Otherwise, use a RSA_AES envelope hybrid
+        # Use the length of the public key to determine the encryption type
+        key_size_bytes = public_key.key_size // 8
+        max_rsa_length = key_size_bytes - 66  # OAEP SHA256 overhead
+        print(max_rsa_length)
```
CyMule, line 617:
```suggestion
```
nit

_test_unstructured_client/unit/test_encryption.py
```diff
@@ -0,0 +1,108 @@
...
+    ("A" * (max_payload_size + 1), "rsa_aes"),  # Just over the RSA limit
+    ("A" * 500, "rsa_aes"),  # Well over the RSA limit
+])
+def test_encrypt_around_rsa_size_limit(rsa_key_pair, plaintext, secret_type):
+    """
+    Test that payloads around the RSA size limit choose the correct algorithm.
+    """
+    _, public_key_pem = rsa_key_pair
+
+    print(f"Testing plaintext of length {len(plaintext)} with expected type {secret_type}")
```
CyMule, line 95:
```suggestion
```
nit

CyMule (approved):

PR #280, opened 2025-06-27
feat: improve PDF validation error handling with FileValidationError …

CyMule:
…base class

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "unstructured-client"
-version = "0.37.2"
+version = "0.37.3"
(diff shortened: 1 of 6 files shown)


mpolomdeepsense (approved):
Code looks good! Not sure why CI fails, there are some ReadTimeout errors so might have been random.

pyproject.toml
awalker4, line 1: This file is autogenerated - we can manually set the version in [gen.yaml](https://github.com/Unstructured-IO/unstructured-python-client/blob/dbf1c073539e5da4cd6721363498b285111ab071/gen.yaml#L17), and then it should take this the next time the SDK is rebuilt.

PR #299, opened 2025-07-24
chore: Manually regenerate SDK

awalker4:
We have a generate failure because we've pulled `general.py` out for some custom changes. (See https://github.com/Unstructured-IO/unstructured-python-client/pull/270) This will hit the occasional bump as new code is added to this file upstream. The steps for fixing this:

- Remove `general.py` from .genignore
- Run speakeasy generate
- Stage and commit the autogenerated changes, working around our custom code
- Commit without the `.genignore` change

### DIFF
--- a/src/unstructured_client/general.py
+++ b/src/unstructured_client/general.py
@@ -7,7 +7,7 @@ from unstructured_client import utils
 from unstructured_client._hooks import HookContext
 from unstructured_client.models import errors, operations, shared
 from unstructured_client.types import BaseModel, OptionalNullable, UNSET
-from unstructured_client._hooks.custom.clean_server_url_hook import clean_server_url
+from unstructured_client.utils.unmarshal_json_response import unmarshal_json_response
 
 
 class PartitionAcceptEnum(str, Enum):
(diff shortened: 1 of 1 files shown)


awalker4:
Uh oh, looks like we need a bit more work to enable it over here - https://github.com/Unstructured-IO/unstructured-python-client/actions/runs/16507745390/job/46682622133

In any case since most of this is autogenerated it may not be very helpful.

yuming-long:
hmm yeah i thought i enabled this already - i will follow up when crag is back :)

yuming-long (approved):
LGTM!

src/unstructured_client/models/errors/unstructuredclienterror.py
```diff
@@ -0,0 +1,26 @@
...
+import httpx
+from typing import Optional
+
+
+class UnstructuredClientError(Exception):
+    """The base class for all HTTP error responses."""
+
+    message: str
+    status_code: int
+    body: str
```
yuming-long, line 12: fyi we raise body back to plugins as `httpexception(detail=e.body)`

yuming-long: and i think the message here could be something like
`"API error occurred: Status 402 Content-Type application/json. Body: {\"detail\": \"api key Not found etc\"}"`

awalker4: Yeah, I would have preferred it if they didn't concatenate everything like that. At least `body` will still be the actual json so we can pull out the server response.

PR #340, opened 2026-05-01
fix(DTPL-364): Fix pdf_split_hook issues

PastelStorm:
> [!NOTE]
> **Medium Risk**
> Changes core request/hook dispatch and split-PDF async execution/cancellation paths, which could affect concurrency behavior and cleanup semantics under load. Added tests mitigate risk, but regressions could surface in real-world event-loop/thread-local hook usage.
> 
> **Overview**
> **Fixes `partition_async()` split-PDF execution to be truly async.** Split chunk collection now runs via awaited async hook dispatch rather than spinning up a nested event loop in a worker thread, with response reassembly offloaded so it doesn’t block the event loop.
> 
> **Adds robust cancellation and resource cleanup.** Introduces async hook dispatch in `SDKHooks` (sync hooks run in `to_thread` with per-hook serialization), adds cancellation-aware cleanup in `BaseSDK.do_request_async`, and hardens `SplitPdfHook` with process-wide PDF-split setup locking, lazy sync executor creation, chunk-task cancellation/draining, and cleanup of tempdirs/unconsumed chunk files. Extensive new unit tests cover ordering, concurrency limits, cancellation, strict-failure behavior, and non-blocking guarantees.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da0352c716ea1cd51aafac1ad4ff63c48fb118d3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aballman (commented):
Reviewed alongside an adversarial pass with codex. Three concerns worth addressing before merge.

**P1 - Cross-thread race in `before_request_async`** (`src/unstructured_client/_hooks/sdkhooks.py:74-79`)

Sync hooks are now offloaded with `asyncio.to_thread()`. With overlapping `partition_async()` calls on the same SDK instance, `SplitPdfHook.before_request()` ends up running concurrently on multiple OS threads while mutating shared instance state. The clearest case is `self.partition_base_url` at `split_pdf_hook.py:536` - two concurrent partition calls with different base URLs can race, and one will send its split orchestration to the wrong host. Same hazard exists for the various per-operation dicts. We should serialize sync-hook execution per SDK instance (e.g. an `asyncio.Lock` in `SDKHooks`) or, better, make the split-PDF state local rather than instance-attached.

**P2 - Cancellation regresses on the heaviest preprocessing path** (`src/unstructured_client/basesdk.py:334-345`)

`asyncio.shield(before_request_task)` deliberately waits for the offloaded sync hook to finish before re-raising `CancelledError`. Because the sync `SplitPdfHook.before_request()` is now in a worker thread and that's where PDF splitting and tempfile writing happen, cancelling `partition_async()` mid-split won't take effect until the thread returns. The new cleanup logic only covers cancellation after setup completes, so this path is actually less responsive to cancellation than before for large PDFs.

**P2 - CancelledError wrapped before reaching `after_error_async`** (`src/unstructured_client/basesdk.py:22-26`)

`cleanup_cancelled_request()` wraps `asyncio.CancelledError` in `_RequestBoundCancelledError` before dispatching to `after_error_async`. Downstream async hooks can no longer detect cancellations via `isinstance(error, asyncio.CancelledError)`, so user-initiated cancels look like ordinary failures. This is a contract break in the new first-class async hook surface and worth fixing now rather than later.

**Nits**

- `_request_task_cancellation` (`split_pdf_hook.py:278-281`) cancels every non-done task on the loop via `asyncio.all_tasks()`. It's safe today because the loop is private to the worker thread, but worth a comment to flag the assumption.
- `del response._request` at `split_pdf_hook.py:760` reaches into an httpx private attribute. Fragile across httpx upgrades.
- No test exercises the P1 cross-thread race. The cancellation-cleanup test added at `test_split_pdf_hook.py` covers the post-setup case but not the during-setup case described in P2.

The async-hook refactor is directionally correct and fixes the nested-event-loop issue, but P1 in particular is a real regression for any SDK instance reused across concurrent `partition_async()` calls.

aballman:
Combined findings from a manual review and an adversarial codex pass. The codex pass did not identify any concrete introduced bug; the items below come from the manual review and from validated Cursor Bugbot findings.

### Medium

**1. `remaining_tasks` count always logs zero on caller cancellation** — `split_pdf_hook.py:170-176`

The count is computed after `await asyncio.gather(*tasks, return_exceptions=True)`, which guarantees every task is `done()`, so the metric will always be `0`. Capture the count before the gather, the way `_collect_task_responses` already does at lines 239 and 254.

**2. `cleanup_cancelled_request` swallows secondary cancellation silently** — `basesdk.py:325-327`

The handler catches `Exception` but not `BaseException`. If the cleanup itself is cancelled mid-await (loop shutdown, parent task cancellation propagation), the `CancelledError` escapes uncaught and split-PDF state can leak. Either widen the catch to `BaseException` with a debug log, or wrap the `after_error_async` call in `asyncio.shield(...)` so cleanup completes before propagating.

### Low

**3. Claimed TOCTOU between `_build_after_success_response` and `_clear_operation`** — `split_pdf_hook.py:1276-1308, 1373-1384`

Cursor Bugbot flagged this; I think it's a false positive. `_clear_operation` runs in the `finally` of `after_success_async` only after `await asyncio.to_thread(_build_after_success_response, ...)` returns, so the two are sequenced rather than concurrent. Worth a second read by someone closer to the code, but I don't see the race.

**4. Class-level `threading.local` shared across instances is non-obvious** — `split_pdf_hook.py:347-355`

`_split_pdf_setup_state = threading.local()` is correct (per-thread state used to assert pypdfium calls happen under the lock), but it's surprising as a class attribute. A one-line comment would help future readers.

**5. `_acquire_split_pdf_setup_slot` busy-polls the bounded semaphore** — `split_pdf_hook.py:765-768`

`while not gate.acquire(blocking=False): await asyncio.sleep(0.01)` adds up to 10ms latency per concurrent caller. An `asyncio.Semaphore` would be both cleaner and lower-latency since admission is already async-only.

**6. Operation-id reuse can clobber in-flight state**

The dicts keyed by `operation_id` would silently overwrite each other if two concurrent `partition_async` calls collide on the same id. Not introduced here, but the surface area expands. A defensive `if operation_id in self.coroutines_to_execute: raise ...` in `_before_request_unlocked` would catch it.

### Nit

**7. `_RequestBoundCancelledError.request` is set but never consumed** — `basesdk.py:21-26`

No after-error hook in this PR reads `error.request`. Either drop the attachment until something uses it, or document the contract for hook authors.

**8. `_finish_cancelled_split_setup` swallows non-cancellation exceptions** — `split_pdf_hook.py:773-784`

Returns `None` after a debug log, which is probably correct (setup never reached the state-mutation point), but the contract is implicit. A short docstring would help.

**9. Test file mislabelled as integration** — `_test_unstructured_client/integration/test_split_pdf_async_behavior.py`

These use `httpx.MockTransport` and have no real backend dependency. Move under `unit/` or rename so the boundary is obvious.

### Informational

**10. Behavior change for existing async callers**

Sync hooks now run on a worker thread under `partition_async`. Anything relying on `contextvars` or thread-local state from the loop thread will silently break. Worth a release-note callout.

**11. Process-wide pypdfium setup serialization**

Setup is now bounded by a single worker thread across all callers. Likely fine because chunk execution dominates, but `partition_async` throughput on many small PDFs is now bounded by setup latency. Worth measuring if benchmarks exist.

aballman (approved):
Approved. Please resolve the two Medium findings (remaining_tasks log and cleanup_cancelled_request BaseException catch) before merging.

PR #341, opened 2026-05-07
feat(PLU-354): remove connector config models from the SDK

awalker4:
## Summary
- Replace the typed unions in `CreateSourceConnectorConfig`, `CreateDestinationConnectorConfig`, `UpdateSourceConnectorConfig`, `UpdateDestinationConnectorConfig`, and the `*ConnectorInformation` configs with `Dict[str, Any]`.
- Delete the per-connector config models and their docs (e.g. `S3SourceConnectorConfig`, `AzureDestinationConnectorConfig`, `OpenSearchConnectorConfig`, …).
- Bump version to 0.44.0 and add a breaking-change changelog entry.

This decouples the SDK from backend connector schemas — adding/removing fields on a connector no longer requires an SDK release.

These were deprecated in our docs last year. None of our SDK snippets use them.

## Test plan
- [ ] CI green
- [ ] Smoke: create a source connector with a plain dict config against SND and confirm the request goes through

---

> [!NOTE]
> **Medium Risk**
> This is a breaking change to the SDK surface area: connector `config` fields switch from typed models/unions to `Dict[str, Any]`, and many generated connector config classes/docs are removed, which may break downstream type checks and runtime imports.
> 
> **Overview**
> **Decouples connector configuration from generated SDK models.** Connector `config` fields for create/update and `*ConnectorInformation` responses now use `Dict[str, Any]` instead of per-connector typed unions/models.
> 
> Removes the generated connector config model classes and their documentation, updates contract tests to assert `config` is a `dict`, and bumps the SDK version to `0.44.0` with a breaking-change entry in `CHANGELOG.md` (plus release metadata in `RELEASES.md`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb0ae929ec7227a3c48114928f5523ce7ce3a39e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/models/shared/updatesourceconnector.md
```diff
@@ -5,4 +5,4 @@
 
 | Field                                                                                    | Type                                                                                     | Required                                                                                 | Description                                                                              |
 | ---------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------- |
-| `config`                                                                                 | [shared.UpdateSourceConnectorConfig](../../models/shared/updatesourceconnectorconfig.md) | :heavy_check_mark:                                                                       | N/A                                                                                      |
```
jordan-homan, line 8: why are most of these getting deleted but some staying? guessing the connector-specific ones are being deleted but not all?

src/unstructured_client/models/shared/astradbconnectorconfig.py
```diff
@@ -1,83 +0,0 @@
-"""Code generated by Speakeasy (https://speakeasy.com). DO NOT EDIT."""
```
jordan-homan, line 1: are all of these deletes a breaking change in any way?  Is there a chance that customers are using these files on their side that will now break if they upgrade the package?

jordan-homan (approved):

src/unstructured_client/models/shared/astradbconnectorconfig.py
awalker4: Totally! We've had a [deprecation notice](https://docs.unstructured.io/api-reference/workflow/migration#connectorconfiginput-and-connectortype-classes-are-deprecated) since like September. So none of our live examples use these classes. If someone has a real old client, hopefully they'll find that migration guide. I'll make a docs pr to note that they're fully removed once this goes out.

docs/models/shared/updatesourceconnector.md
awalker4: Yep, just the generic classes should stay, and they'll just wrap a dict. So we may be able to clean those up too longer term.

PR #344, opened 2026-06-04
feat: env-var configurable httpx pool + TLS in split_pdf_hook (0.45.0)

fxdgear:
## What

Adds env-var knobs for the `httpx.AsyncClient` used by `split_pdf_hook.run_tasks`, and ships them as **0.45.0**. Defaults match httpx — fully backward compatible.

### Connection-pool limits
- `UNSTRUCTURED_CLIENT_MAX_CONNECTIONS` (default `100`)
- `UNSTRUCTURED_CLIENT_MAX_KEEPALIVE_CONNECTIONS` (default `20`)
- `UNSTRUCTURED_CLIENT_KEEPALIVE_EXPIRY` (default `5.0` seconds)

### TLS trust store (server verification)
Honors the standard env vars other Python tooling already respects, so a single setting applies uniformly:
- `SSL_CERT_FILE` (stdlib `ssl` convention)
- `REQUESTS_CA_BUNDLE` (requests / httpx-ecosystem convention; used if `SSL_CERT_FILE` is unset)

### mTLS client certificate
- `UNSTRUCTURED_CLIENT_TLS_CLIENT_CERT` — PEM file (httpx reads key from the same file by default)
- `UNSTRUCTURED_CLIENT_TLS_CLIENT_KEY` — optional, when cert and key live in separate files

### Observability
- Extends the existing `split_pdf event=plan_created` INFO log to include the resolved pool values and trust-store / mTLS mode, so the active config is visible in production logs without leaking filesystem paths.

### Release
- Bumps `_version.py` to `0.45.0`, adds a `0.45.0` `CHANGELOG.md` section, and appends a matching `RELEASES.md` entry.

## Why

When the SDK runs in an environment where load balancing happens at TCP-connect time rather than per-request (a common Kubernetes setup with a plain ClusterIP and no service mesh), httpx's default keepalive pooling can lock onto a subset of backends. Newly added backends never receive traffic because existing connections stay glued to the originally-resolved set.

Letting operators force shorter keepalive (e.g. `MAX_KEEPALIVE_CONNECTIONS=1` + a low `KEEPALIVE_EXPIRY`) makes the client re-establish connections more frequently, redistributing across the available backends.

The TLS additions are for SDK consumers running behind corporate proxies with custom CAs, or against backends that require mTLS — previously they had to subclass / monkey-patch to get a custom `verify` or `cert` into the split-PDF client.

## How to use

```yaml
env:
  # Pool reshuffling for connect-time-only LBs
  - name: UNSTRUCTURED_CLIENT_MAX_KEEPALIVE_CONNECTIONS
    value: "1"
  - name: UNSTRUCTURED_CLIENT_KEEPALIVE_EXPIRY
    value: "30.0"

  # Custom trust store (standard env var, picked up by httpx, requests, ssl)
  - name: SSL_CERT_FILE
    value: /etc/ssl/internal-ca-bundle.pem

  # mTLS
  - name: UNSTRUCTURED_CLIENT_TLS_CLIENT_CERT
    value: /etc/ssl/client.crt
  - name: UNSTRUCTURED_CLIENT_TLS_CLIENT_KEY
    value: /etc/ssl/client.key
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 0.45.0
+
+### Features
+* Make the split-PDF `httpx.AsyncClient` connection-pool limits configurable via env vars: `UNSTRUCTURED_CLIENT_MAX_CONNECTIONS` (default `100`), `UNSTRUCTURED_CLIENT_MAX_KEEPALIVE_CONNECTIONS` (default `20`), and `UNSTRUCTURED_CLIENT_KEEPALIVE_EXPIRY` (default `5.0`s). Defaults match httpx, so behavior is unchanged unless set. Useful when deploying behind a connect-time-only load balancer (e.g. Kubernetes ClusterIP without a mesh) where shorter keepalives force connections to redistribute across backend pods.
+* Honor the standard `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE` env vars to point the split-PDF `httpx.AsyncClient` at a custom trust store, so a single env-var setting applies uniformly across Python tooling.
+* Add `UNSTRUCTURED_CLIENT_TLS_CLIENT_CERT` and `UNSTRUCTURED_CLIENT_TLS_CLIENT_KEY` env vars to wire an mTLS client certificate into the split-PDF `httpx.AsyncClient` (single PEM, or split cert + key files).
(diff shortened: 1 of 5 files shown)


ctrahey (commented):
Please update every httpx client to accept configurable TLS certificates as we touch them!

awalker4 (approved):
LGTM! The repo is no longer auto-generated, so we're empowered to own the changelog and bump the version before doing a github release.

